PR #92, opened 2025-08-06
Crate: WIP using `tiny-skia` to render glyphs

kane50613:
Dropping `swash` and replace our own implementation with `tiny-skia` for glyph rendering, reduces WASM size from 2.8MB down to 2.3MB

But the gzipped size doesn't seems to change by a lot (~1.2MB after gzip with example/cloudflare-workers)

Not the top priority so putting it in draft state.


kane50613:
zeno can now do the job #150 , closing this

PR #530, opened 2026-03-07
fix(background): support background-size auto

relufi:
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/background-size

## If the background-size is auto or auto auto:

- If the image has both horizontal and vertical intrinsic dimensions, it's rendered at that size.
- If the image has no intrinsic dimensions and has no intrinsic proportions, it's rendered at the size of the background positioning area.
- If the image has no intrinsic dimensions but has intrinsic proportions, it's rendered as if contain had been specified instead.
- If the image has only one intrinsic dimension and has intrinsic proportions, it's rendered at the size corresponding to that one dimension. The other dimension is computed using the specified dimension and the intrinsic proportions.
- If the image has only one intrinsic dimension but has no intrinsic proportions, it's rendered using the specified dimension and the other dimension of the background positioning area.
## If the background-size has one auto component and one non-auto component:
- If the image has intrinsic proportions, it's stretched to the specified dimension. The unspecified dimension is computed using the specified dimension and the intrinsic proportions.
- If the image has no intrinsic proportions, it's stretched to the specified dimension. The unspecified dimension is computed using the image's corresponding intrinsic dimension, if there is one. If there is no such intrinsic dimension, it becomes the corresponding dimension of the background positioning area.

## Summary by CodeRabbit

* **Bug Fixes**
  * Background sizing now uses image intrinsic dimensions when width or height is set to Auto, preserving aspect ratio.
  * If both dimensions are Auto and intrinsic size is available, the intrinsic size is used.
  * When intrinsic size is zero or image cannot be resolved, previous area-based fallback is used.
  * Cover behavior unchanged, still prefers intrinsic dimensions when available.

### DIFF
--- a/.changeset/fix-background-position
+++ b/.changeset/fix-background-position
@@ -1,0 +1,5 @@
+---
+"takumi": patch
+---
+
+Fix `background-position`, `mask-position` now default to `top left`


takumi/src/rendering/background_drawing.rs
```diff
@@ -159,10 +159,7 @@ impl BackgroundTile {
 }
 
 pub(crate) fn resolve_length_against_area(unit: Length, area: u32, sizing: &Sizing) -> u32 {
```
kane50613: this function can be inined if just one line.

PR #658, opened 2026-04-13
refactor(helpers): update twemoji to v17

SerenModz21:
Twemoji was forked by the original team that created it (before being laid off) and has been getting maintained since then.

As such, I have updated the link for the Twemoji SVGs.

See here for more details: https://github.com/jdecked/twemoji

> [!IMPORTANT]
> The changeset CLI (`bunx changeset`) thinks this is a major bump. However, I believe this to be incorrect.
>
> I have mentioned this below also: https://github.com/kane50613/takumi/pull/658#discussion_r3073581942

## Summary by CodeRabbit

* **Chores**
  * Updated emoji asset source to Twemoji v17.0.2 via a new CDN path for improved coverage and stability.
  * Added a changeset recording a patch release that updates Twemoji to v17.0.2.
* **Tests**
  * Updated emoji-related test expectations to reference the new Twemoji v17.0.2 asset URLs.

### DIFF
--- a/.changeset/vast-places-mix.md
+++ b/.changeset/vast-places-mix.md
@@ -1,5 +1,5 @@
 ---
-"@takumi-rs/helpers": major
+"@takumi-rs/helpers": patch
 ---
 
 Update Twemoji to v17
--- a/.changeset/vast-places-mix.md
+++ b/.changeset/vast-places-mix.md
@@ -1,0 +1,5 @@
+---
+"@takumi-rs/helpers": major
+---
+
+Update Twemoji to v17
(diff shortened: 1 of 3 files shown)


.changeset/vast-places-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@takumi-rs/helpers": major
```
SerenModz21: I have never used changesets before, so I don't know if I could just change it. However, I don't think it should be a major change.

SerenModz21: I think this should actually be a patch rather than a major bump.

Will wait for feedback before proceeding.

kane50613: it's ok I'll help fix it

SerenModz21: Thank you very much! My apologies for missing the emoji tests.

kane50613: nooooo dont be sorry thanks for catching the twemoji version 😁

PR #717, opened 2026-05-22
NAPI: fix broken type definitions for @takumi-rs/core

### LINKED ISSUE #715: @takumi-rs/core types are broken?
### Pre-submission Checklist

### Description

Looks like the `export.d.ts` tries to `import { Renderer as NativeRenderer } from "../index";` but there is no `../index` file in the distribution, breaking the type definition for `Renderer`. Runtime is unaffected, only the type definition. 

### Steps to Reproduce

1. `import { Renderer } from "@takumi-rs/core";`
2. create a Renderer instance with options (like fonts).
3. type check the code.
4. observe the Renderer instance type cannot be resolved.

### Expected Behavior

Types should resolve. This worked in version `1.1.2`, but broke with `1.2.1` and is still broken in `1.3.0`.

### Actual Behavior

Renderer type is not resolved.

### Reproduction Code

```

const fonts = [
  ...(fonts data),
];

const renderer = new Renderer({
  fonts,
});
```

Type check the above and Renderer will show as not excepting arguments because it has no type definition.

### Affected Package

@takumi-rs/core

### Version

1.3.0

### Operating System

macOS

### Environment Details

_No response_

### Playground Link

_No response_

### Additional Context

_No response_

kane50613:
## Summary

Fixes #715.

`dist/export.d.ts` was emitted with `import { Renderer } from \"../index\"`, but `index.d.ts` wasn't in the published `files` array — so consumers saw `Renderer` as `any`. Two changes restore the types:

1. Add `index.d.ts` to `files` so the napi-generated declarations ship with the package.
2. Change the source imports to `../index.js` so node16 ESM module resolution can find the file (extensionless paths fail in strict ESM resolution).

## Guard

A new `verify-napi-pkg` CI job runs `publint` + `@arethetypeswrong/core` against the packed tarball and would have caught this regression: it reported `InternalResolutionError` for `../index` across all three resolution modes (node10, node16, bundler) when reverted to the broken state.

The job is wired into `pass-test` so it blocks merge.

## Test plan

- [x] `bun run verify:pkg` passes on the fix
- [x] `bun run verify:pkg` fails on the pre-fix state (verified locally)
- [x] `bun test ./tests/` in `takumi-napi-core` — 55 pass, 0 fail

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed broken type definitions for `@takumi-rs/core`.

* **Chores**
  * Updated module specifiers for ESM compatibility.
  * Added automated package verification to CI pipeline to validate published package contents and types.
  * Added type checking tools to development dependencies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kane50613/takumi/pull/717?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/napi-core-publish-index-dts.md
+++ b/.changeset/napi-core-publish-index-dts.md
@@ -1,0 +1,5 @@
+---
+"@takumi-rs/core": patch
+---
+
+Fix broken type definitions for `@takumi-rs/core` (#715)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -527,6 +527,37 @@ jobs:
           docker run --rm -v $PWD:/workspace -w /workspace/takumi-napi-core \
             oven/bun:alpine bun test
 
+  verify-napi-pkg:
+    name: Verify @takumi-rs/core package
+    needs: [build-napi, build-helpers]
+    runs-on: ubuntu-slim
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@main
+        with:
+          bun-version: latest
+
+      - name: Download helpers artifact
+        uses: actions/download-artifact@v8
+        with:
+          name: helpers-dist
+          path: takumi-helpers/dist
+
+      - name: Download N-API bindings artifact
+        uses: actions/download-artifact@v8
+        with:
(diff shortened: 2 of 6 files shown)


kane50613:
Closing — will make the minimal one-line fix directly.

PR #718, opened 2026-05-23
Crate: fix `-webkit-text-stroke` width with `text-fit`

kane50613:
## Summary

`-webkit-text-stroke-width` was being scaled by the `text-fit` scale factor because stroke is applied to glyph paths in path-local space *before* the text-fit transform in `render_mask`. With `grow` (and large scale factors) a 12px stroke ballooned to ~60px; with `background-clip: text` the stroke mask also exceeded the rasterized clip-image bounds and disappeared.

## Fix

Divide `stroke_width` by the transform's uniform scale at the call sites (`draw_text_stroke`, `draw_text_stroke_clip_image`), so the painted stroke stays in CSS pixels regardless of text-fit scale. For non-text-fit text the transform is translation-only, so `uniform_scale = 1.0` and behavior is preserved.

## Chrome alignment

This matches Chrome's `kFontSize` text-fit method ([`third_party/blink/renderer/core/style/fit_text.cc`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/style/fit_text.cc;l=13-17)): in `kFontSize` mode the text is reshaped at the adjusted font-size, so `webkit-text-stroke-width` stays at its CSS-pixel value. The cheap compensation here produces the same visual result without paying the cost of a re-shape pass (matters for perf and binary size).

## Test plan

- [x] New fixture `text_fit_with_webkit_text_stroke` (grow / grow 150% / shrink / shrink 70%)
- [x] All 188 existing fixture tests still pass
- [x] `cargo clippy` clean

## Release Notes

* **Bug Fixes**
  * Fixed `-webkit-text-stroke-width` scaling behavior with `text-fit`. Stroke width now applies in CSS pixels regardless of the text-fit scale factor, aligning with standard browser rendering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kane50613/takumi/pull/718?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


kane50613:
Reopening with a proper Quote.tsx-specific reproduction and fix.

PR #735, opened 2026-05-29
Crate: accept inline image bytes for image node `src`

kane50613:
`ImageNode.src` is typed `string | Uint8Array | ArrayBuffer`, but inline bytes never actually rendered — and not because of `skip_deserializing` (that's on the unrelated `Loaded` variant). The real cause: `ImageSourceInput::Buffer` was `Vec<u8>`, which deserializes via `deserialize_seq` and only accepts a JSON-style array of numbers. A real `Uint8Array`/`ArrayBuffer` crosses both the napi and WASM boundaries as a *bytes value* (`visit_bytes` / `visit_byte_buf`), so it never matched the untagged enum and deserialization failed on **both** `@takumi-rs/core` and `@takumi-rs/wasm`.

This replaces the original JS-side base64 workaround with a one-type Rust fix: `Buffer(Vec<u8>)` → `Buffer(serde_bytes::ByteBuf)`. `ByteBuf` deserializes via `deserialize_bytes`, which serde's untagged machinery satisfies from buffered `Content::Bytes`/`ByteBuf`. Inline raster and SVG bytes are decoded directly by the renderer (`ImageSource::from_bytes` sniffs the format) — no base64 round-trip, ~33% smaller payloads.

### Behaviour
- `image({ src })` accepts `Uint8Array` / `ArrayBuffer` directly; the helper passes `src` through unchanged.
- Raster bytes (invalid UTF-8) → `Buffer` variant → format-sniffed. SVG markup (valid UTF-8) → `Url` variant → existing SVG path. No JS special-casing needed.
- **Limitation:** bytes ride the native/WASM boundary as-is. A node tree explicitly serialized to JSON (e.g. posted to the HTTP server) still needs a `data:` URI string.

### Tests
- Core unit test deserializing a bytes value into `Buffer` (red on `Vec<u8>`, green on `ByteBuf`).
- napi integration: inline `Uint8Array`/`ArrayBuffer` and SVG-as-bytes render pixel-identical to the fetchedResources / string paths.
- wasm verified by mechanism (identical `Content::ByteBuf` path) + compile + clippy.

## Summary by CodeRabbit

* **New Features**
  * image() now accepts inline byte arrays (Uint8Array/ArrayBuffer) as the image source.
  * Inline SVG content can be provided either as plain text or as encoded bytes.

* **Tests**
  * Added render tests validating that inline image bytes and encoded SVG inputs render successfully and produce matching outputs to existing references.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kane50613/takumi/pull/735?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


kane50613:
@coderabbitai review

PR #745, opened 2026-06-06
Crate: move `SizingContext` into `layout::style`

kane50613:
Relocates the length-resolution context (renamed `Sizing` → `SizingContext`) from `rendering` into `layout::style`.

It only depends on style types (`CalcArena`, `Viewport`), so this removes the `style → rendering` import cycle — a prerequisite for extracting the CSS layer into its own crate (`takumi-css`) where the cold CSS parser can be size-optimized independently. Pure move + rename — no behavior change.

## Release Notes

* **Refactor**
  * Reorganized internal layout sizing infrastructure to improve code maintainability and separation of concerns across styling, animation, and rendering systems.

PR #746, opened 2026-06-06
Crate: move `fast_div_255` to `layout::style`

kane50613:
Stacked on #745.

`fast_div_255`/`fast_div_255_u32` are pure primitives needed by value-level colour math (`color-mix`), which must stay in `style`. Moves them into `layout::style` (`rendering` re-exports for its own use). Removes `color.rs`'s last `rendering` dependency.

First step of lifting the rasterization layer out of `style` (see plan in #745). Pure move — no behavior change.


kane50613:

PR #747, opened 2026-06-06
Crate: narrow gradient rasterization to `SizingContext`

kane50613:
Stacked on #746.

The three `*GradientTile::new` constructors and `resolve_stops_along_axis` took a full `&RenderContext` but only read `sizing` + `current_color`. Pass those directly (`&SizingContext`, `Color`) so the gradient value modules no longer depend on `rendering` — cutting the bulk of the `style → rendering` cycle.

Adds `SizingContext::new_test` for tests. One overlay reference test still uses `blend_pixel`; it relocates with the hot overlay loop in a follow-up (that code belongs in `rendering`, not the soon-to-be size-optimized CSS crate). No behavior change.


kane50613:

PR #748, opened 2026-06-06
Crate: narrow stylesheet-animation sampling to `SizingContext`

kane50613:
Stacked on #747. `apply_stylesheet_animations` only read `stylesheet/time/sizing/current_color` from its `&RenderContext`; pass those directly so `animation.rs` no longer depends on `rendering`. No behavior change.


kane50613:

PR #749, opened 2026-06-06
Crate: move clip-path mask rasterization to rendering

kane50613:
Stacked on #748. `BasicShape::render_mask` builds paths and calls the rendering `render_mask` — genuine rasterization. Moves it to `rendering::canvas::mask` as `render_clip_shape_mask` (with `resolve_radius` + `From<FillRule> for Fill`). Shape value types + `FromCss`/`ToCss` stay in `style`, which no longer depends on `rendering` for clip-path. No behavior change.


kane50613:

PR #752, opened 2026-06-06
Crate: extract the CSS layer into a `takumi-css` crate

kane50613:
Stacked on #751. Moves the cold CSS layer (`layout::style`, `Viewport`, `keyframes`, style errors) into a new internal `takumi-css` crate so it can be size-optimized independently of the hot rendering paths.

- **Selector matching stays on the takumi side** (`layout::matching`) — it depends on the node tree (`Node`), so it doesn't belong in the pure-CSS crate. `takumi-css` has zero node/render dependencies.
- `takumi` re-exports `takumi_css::style`/`Viewport`/`keyframes`/errors, so all existing paths resolve unchanged.
- `#[non_exhaustive]` dropped from the CSS value types (the renderer must exhaustively handle every variant across the crate boundary).
- Internal companion crate, versioned in lockstep with `takumi`; CI dry-run publishes both in dependency order.
- Prunes now-unused takumi deps (bitflags, paste, phf, precomputed-hash).

Verified: full workspace + wasm build, `clippy --all-targets --all-features -D warnings` clean, `cargo machete` clean, tests pass, publish dry-run of both crates succeeds. No behavior change. Sets up the final PR: `opt-level="s"` on `takumi-css`.


kane50613:
Consolidated into #753 (extract + opt-level now one PR).

Issue #728, opened 2026-05-29
V2 Breaking Changes Checklist / Roadmap


kane50613:
Breaking changes for v2, from auditing real usage (445 GitHub sites, 62 source reads). Bundle them into one major bump so users migrate once. Non-breaking improvements ship separately on 1.x.

## CSS spec conformance

Deviations from CSSWG initial values, inheritance, and value semantics, found while auditing `takumi/src/layout/style/`. Split into bugs vs. deliberate next/og-compat choices to ratify or align.

### Bugs (wrong rendered result)

- [x] **`border-width` / `outline-width` default to `0`**: omitted width resolves to `0` (`Length::Auto`->0), so `border: solid red` and `outline: solid` render nothing. CSS initial is `medium` (~3px). (`length.rs` `Default for Length`; `border.rs`; `stylesheets.rs` `to_taffy_style`)
- [x] **Negative `scale` collapses to `0`**: `PercentageNumber::from_css` clamps `.max(0.0)` (`percentage_number.rs:73`), so `scale: -1` and `transform: scaleX(-1)`/`scaleY(-1)`/`scale(-1)` (`transform.rs:527-541`) zero out instead of reflecting. Allow negative factors.
- [x] **`line-clamp` inherits**: `line_clamp: ... where inherit = true` (`stylesheets.rs:997`). CSS `line-clamp` is a *shorthand* (`max-lines` + `block-ellipsis` + `continue`) whose longhands have **mixed** inheritance (only `block-ellipsis` inherits); takumi's collapsed model mirrors legacy `-webkit-line-clamp`, which is **not** inherited (MDN). A single `inherit` flag can't cover that, so drop it (the `count`/max-lines part must not inherit).
- [x] **`-webkit-line-clamp` / `WebkitLineClamp` is dropped**: `webkit_property_id_from_name` (`stylesheets_helpers.rs:214`) only maps `text-stroke*` / `text-fill-color`, so the webkit-prefixed line-clamp normalizes to `webkit_line_clamp` -> `Ignored`. Only bare `lineClamp`/`line-clamp` and Tailwind `line-clamp-*` work, but next/og users write `WebkitLineClamp`. Add the alias.

### Cleanup (same shorthand/inheritance footgun)

[...]

PR #768, opened 2026-06-12
Crate: Fix vendor-prefixed property names not being resolved

mvanhorn:
## Summary

`-webkit-line-clamp` and `WebkitLineClamp` now resolve to the registered `line-clamp` longhand instead of normalizing to `Ignored`. Users migrating from next/og (Satori) write `WebkitLineClamp` in style objects, and their clamping silently did nothing because `webkit_property_id_from_name` only mapped the `text-stroke*` and `text-fill-color` suffixes.

## Why this matters

This is the "-webkit-line-clamp silently dropped, add the alias" item from your v2 audit checklist in #728. No call-site wiring is needed: `PropertyId::from_name` already consults `webkit_property_id_from_name` before normalization, so the alias covers both kebab-case CSS and camelCase style-object inputs.

## Testing

`property_id_accepts_webkit_aliases` in `stylesheets_tests.rs` now asserts `-webkit-line-clamp`, `WebkitLineClamp`, and `WebKitLineClamp` resolve to the line-clamp property and not `Ignored`; the takumi-css suite passes locally (471 tests). A patch changeset is included per CONTRIBUTING.md.

Refs #728

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed recognition of vendor-prefixed and variant property names so they resolve to the standard line-clamp property.

* **New Features**
  * Added normalization to accept WebKit-prefixed, vendor-prefixed, and camel-cased property name variants as equivalent to line-clamp.

* **Tests**
  * Expanded tests to verify prefixed and standard line-clamp inputs produce identical behavior.

### DIFF
--- a/.changeset/webkit-line-clamp-alias.md
+++ b/.changeset/webkit-line-clamp-alias.md
@@ -2,4 +2,4 @@
 "takumi-css": patch
 ---
 
-Map `-webkit-line-clamp`, `WebkitLineClamp`, and `WebKitLineClamp` to the existing `line-clamp` property.
+Fix vendor-prefixed property names not resolved properly
--- a/.changeset/webkit-line-clamp-alias.md
+++ b/.changeset/webkit-line-clamp-alias.md
@@ -1,5 +1,6 @@
 ---
+"takumi": patch
 "takumi-css": patch
 ---
 
-Fix vendor-prefixed property names not resolved properly
+Fix vendor-prefixed property names not being resolved
--- a/.changeset/webkit-line-clamp-alias.md
+++ b/.changeset/webkit-line-clamp-alias.md
@@ -1,0 +1,5 @@
+---
+"takumi-css": patch
+---
+
+Map `-webkit-line-clamp`, `WebkitLineClamp`, and `WebKitLineClamp` to the existing `line-clamp` property.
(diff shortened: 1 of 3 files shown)


kane50613:
Thanks for the PR

Sorry I think I've wrote it incorrectly, I was about to fix treading `line-clamp` as a longhand instead it should be shorthand according to [spec](https://drafts.csswg.org/css-overflow-4/#line-clamp) that extends to [max-lines](https://drafts.csswg.org/css-overflow-4/#propdef-max-lines), [block-ellipsis](https://drafts.csswg.org/css-overflow-4/#propdef-block-ellipsis), and [continue](https://drafts.csswg.org/css-overflow-4/#propdef-continue)

which might be a breaking change

I'll stack my simplification on top of this PR to eliminate the need to list all the browser specific prefixes and properties and just redirect to same parser function

mvanhorn:
Thanks for landing this, @kane50613 - vendor-prefixed line-clamp now resolves to the registered longhand instead of getting ignored.

kane50613:
> Thanks for landing this, @kane50613 - vendor-prefixed line-clamp now resolves to the registered longhand instead of getting ignored.

hope to see more chances of Takumi growing!

PR #776, opened 2026-06-14
Default border and outline width to medium

### LINKED ISSUE #728: V2 Breaking Changes Checklist / Roadmap

mvanhorn:
## Summary

When the width token is omitted from the `border` shorthand, `Border::from_css`
in `takumi-css/src/style/properties/border.rs` fell back to
`width.unwrap_or_default()`, and `Length::default()` is `Px(0.0)`. As a result
`border: solid red` and `outline: solid` resolved to a zero width and painted
nothing.

This changes the omitted-width fallback to the CSS initial value `medium`
(~`3px`) when a visible border style is present (`style.is_rendered()`), while
keeping width `0` when no visible style is declared (`border: none`, color-only,
or empty). Because the `outline` shorthand parses through `Border` as well, this
one change fixes both `border` and `outline`.

`Border::default()` is unchanged, so elements with no border declared still
reserve zero width and there is no rendering regression.

## Why this matters

CSS specifies the initial value of `border-width` / `outline-width` as `medium`,
not `0` (https://drafts.csswg.org/css-backgrounds-3/#border-width). The previous
behavior silently dropped any border or outline declared without an explicit
width, which is the first CSS-conformance item called out in the v2 audit
(https://github.com/kane50613/takumi/issues/728).

## Testing

- Added `test_parse_border_style_and_color_defaults_medium_width` (the
  `border: solid red` case from the issue) and updated
  `test_parse_border_style_only` plus the `border-left` style-only assertion in
  `stylesheets_tests.rs` to expect `Px(3.0)`.
- Existing tests for explicit widths, `border: none`, empty, and color-only
  declarations are unchanged and still pass.
- `cargo test -p takumi-css` passes (all tests green).
- Added a changeset (`patch`) for `takumi-css` / `takumi`.

This PR addresses the border / outline-width item from the audit.

Closes #728

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated border and outline width behavior to default to `medium` (~3px) when omitted but a visible style is applied (e.g., `border: solid`), aligning with CSS standard behavior. Non-rendered borders retain zero width defaults.

### DIFF
--- a/.changeset/border-width-default-medium.md
+++ b/.changeset/border-width-default-medium.md
@@ -1,0 +1,6 @@
+---
+"takumi-css": patch
+"takumi": patch
+---
+
+Default an omitted `border-width` / `outline-width` to `medium` (~3px) instead of `0`, so `border: solid red` and `outline: solid` paint a visible line as CSS specifies. Borders with no visible style (`border: none`, color-only, empty) keep width `0`.
--- a/takumi-css/src/style/properties/border.rs
+++ b/takumi-css/src/style/properties/border.rs
@@ -6,6 +6,13 @@ use crate::style::{
   SizingContext, properties::Length,
 };
 
+/// CSS initial value for `border-width` / `outline-width` keyword `medium`.
+///
+/// Per CSS, an omitted width on a visible border/outline resolves to `medium`,
+/// which user agents render as roughly `3px` (not `0`). See
+/// <https://drafts.csswg.org/css-backgrounds-3/#border-width>.
+const MEDIUM_BORDER_WIDTH_PX: f32 = 3.0;
+
 /// Parsed `border` value.
 #[derive(Debug, Default, Clone, Copy, PartialEq)]
 #[non_exhaustive]
@@ -50,9 +57,23 @@ impl<'i> FromCss<'i> for Border {
       ));
     }
 
+    let style = style.unwrap_or_default();
+
+    // When the width token is omitted, CSS resolves it to `medium` (~3px) for a
+    // visible border/outline, so `border: solid red` and `outline: solid` paint
+    // a line. With no visible style (`border: none`, color-only, or empty) the
+    // width stays `0`, keeping `Border::default()` unchanged.
+    let width = width.unwrap_or_else(|| {
+      if style.is_rendered() {
+        Length::Px(MEDIUM_BORDER_WIDTH_PX)
+      } else {
+        Length::default()
+      }
+    });
+
     Ok(Border {
-      width: width.unwrap_or_default(),
-      style: style.unwrap_or_default(),
+      width,
+      style,
       color: color.unwrap_or_default(),
     })
   }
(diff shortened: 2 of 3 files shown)


kane50613:
imo this requires a breaking changes to the underlying `border-width` property to be a new enum with the keywords instead.

One thing I can possibly do is to decouple the versioning of takumi rust crate and takumi-js therefore the js end doesn't require any breaking changes at all.

PR #802, opened 2026-06-21
Align Rust/JS render surface and harden output-format types

kane50613:
Addresses the DX/API findings from the v2 audit (Joshua Bloch's *How to Design a Good API* plus the Rust API Guidelines and TypeScript library best practices).

## Changes

**Naming symmetry (Rust ↔ JS)**
- `measure_layout` → `measure`
- `render_sequence_animation` → `render_animation`
- `ImageOutputFormat` → `OutputFormat`

These now match the JS binding method names so the two surfaces read as one language.

**Hide `image::RgbaImage` from the public API**
- `render` returns a `Bitmap` newtype; `write_image` takes `&Bitmap`.
- Pixels are reachable via `Bitmap::as_raw` / `into_raw`; the public API no longer commits to a specific `image` crate version.

**Discriminated-union output format (napi + wasm)**
- The JS `format`/`quality`/`lossless` flag-bag becomes a discriminated union, so `quality` and `lossless` only appear on the formats that accept them. Impossible combos (e.g. `{ format: "png", quality, lossless }`) no longer typecheck.
- Format strings stay identical across napi and wasm so the universal `takumi-js` layer composes cleanly; wasm exposes WebP as lossless-only (lossy WebP is native-only).

**Quality consistency**
- napi animation/encode paths now clamp out-of-range WebP quality instead of throwing, matching the core `Quality` type and the wasm binding.

## Verification
- `cargo build`, `cargo clippy`, and `cargo test` (903 tests) pass on native and wasm targets.
- `tsc --noEmit` passes for `@takumi-rs/core`, `@takumi-rs/wasm`, and `takumi-js`.
- `oxlint` + `oxfmt` + `rustfmt` clean.

## Notes / follow-ups
- Animation output options (`RenderAnimationOptions`/`EncodeFramesOptions`) still use a flat `format`/`quality`/`lossless` shape; converting those to a union is a natural follow-up.

https://claude.ai/code/session_01PffFyJn2SZ3q1aSdLAyiBb

## Release Notes

* **New Features**
  * JavaScript/WebAssembly output format is now a discriminated union, with `quality` and `lossless` properties only on applicable format variants.

* **Breaking Changes**
  * Rust API: `measure_layout()` → `measure()`, `render_sequence_animation()` → `render_animation()`.
  * `render()` now returns `Bitmap` wrapper type instead of raw image.
  * Quality is now part of `OutputFormat` enum variants rather than a separate function parameter.
  * `write_image()` signature updated.

* **Bug Fixes**
  * Out-of-range WebP quality values are now clamped instead of throwing errors.

### DIFF
--- a/.changeset/js-output-format-union.md
+++ b/.changeset/js-output-format-union.md
@@ -1,0 +1,7 @@
+---
+"@takumi-rs/core": major
+"@takumi-rs/wasm": major
+"takumi-js": major
+---
+
+Model the output format as a discriminated union so `quality` and `lossless` only appear on the formats that accept them, and clamp out-of-range WebP quality instead of throwing
--- a/.changeset/rust-render-naming-bitmap.md
+++ b/.changeset/rust-render-naming-bitmap.md
@@ -1,0 +1,6 @@
+---
+"takumi": major
+"takumi-raster": minor
+---
+
+Rename `measure_layout` to `measure`, `render_sequence_animation` to `render_animation`, and `ImageOutputFormat` to `OutputFormat`; return a `Bitmap` newtype from `render` instead of `image::RgbaImage`, and take `&Bitmap` in `write_image`
--- a/docs/content/docs/upgrade/v2.mdx
+++ b/docs/content/docs/upgrade/v2.mdx
@@ -167,14 +167,29 @@ takumi = { version = "*", default-features = false, features = ["raster-backend"
 
 ### Image output quality is modeled per format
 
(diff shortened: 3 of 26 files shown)


kane50613:
Addressed CodeRabbit's review in 0f984776:

- **`Bitmap` constructor**: did *not* make `from_rgba` `pub` — its parameter is `image::RgbaImage`, so exposing it would re-leak the `image` crate into the public API, which is exactly what the `Bitmap` newtype removes. Instead added `pub fn Bitmap::from_raw(width, height, data: Vec<u8>) -> Option<Self>` (mirrors `RgbaImage::from_raw`), so callers can build frames from their own pixels without depending on `image`. `from_rgba` stays `pub(crate)`.
- **Clamp test**: added `RIFF` / `WEBP` header assertions so it proves the clamped quality still produces valid WebP.

PR #803, opened 2026-06-21
NAPI, WASM: add `renderSvg`

kane50613:
## What

Expose the `takumi-svg` vector backend through both language bindings as a `renderSvg` method that returns an SVG document string.

| Binding | Signature |
| --- | --- |
| `@takumi-rs/core` (napi) | `renderSvg(node, options?, signal?): Promise<string>` |
| `@takumi-rs/wasm` | `renderSvg(node, options?): string` |

## Design

SVG is not modelled as another `OutputFormat` variant: it produces text, not a `Bitmap` encoding, and honours a different option subset. A separate method keeps each return type and option set honest — the most-specific type, with misuse made impossible rather than silently ignored.

`SvgRenderOptions` drops the raster-only knobs: `format`, `quality`, `lossless`, `dithering`, `drawDebugBorder`, `devicePixelRatio` (SVG is vector, so DPR is meaningless). The wasm TS type derives via `Omit<RenderOptions, ...>` to avoid duplicating field docs.

Out of scope: animation (the svg crate has no animation path) and a data-url helper.

## Tests

`renderSvg` parity test per binding: asserts well-formed `<svg>…</svg>` markup, viewport dimensions, and `timeMs`-driven keyframe resolution. Full suites green (napi 51, wasm 49); clippy clean on all three crates.

https://claude.ai/code/session_01PffFyJn2SZ3q1aSdLAyiBb

## Summary by CodeRabbit

* **New Features**
  * Added SVG rendering support with a new `renderSvg()` method on the Renderer class
  * Customize SVG output with the new `SvgRenderOptions` configuration for dimensions, fonts, images, and stylesheets

### DIFF
--- a/.changeset/render-svg-bindings.md
+++ b/.changeset/render-svg-bindings.md
@@ -1,0 +1,6 @@
+---
+"@takumi-rs/core": minor
+"@takumi-rs/wasm": minor
+---
+
+Add `renderSvg` to render a node tree to an SVG document string
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1576,6 +1576,7 @@ dependencies = [
  "serde",
  "takumi-core",
  "takumi-raster",
+ "takumi-svg",
 ]
 
 [[package]]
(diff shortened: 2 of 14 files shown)


kane50613:
@coderabbitai review

PR #910, opened 2026-07-05
Core: seal foreign types out of the public API

kane50613:
## Why

takumi-core ships as its own crate, but its public API leaked `taffy`, `parley`, and `tiny_skia` — anyone who only parses CSS or holds a node tree was dragged into depending on the layout engine's crates, and the surface couldn't reach a clean 2.0. A few properties also carried the wrong CSS initial value.

## What

- Gate the layout and paint engine (`layout::{tree,inline,border}`, `paint`, `scene`, `geometry`, `shadow`) behind an `unstable` feature that the backend crates enable. The default takumi-core API no longer names a foreign type — verified with `cargo public-api`. The ours-to-foreign `From` impls stay, now feature-gated, so no relocated conversion helpers are needed.
- Add `MaxSize` (`none`) and `Gap` (`normal`) value types so `max-width`/`max-height` and the gaps carry their real initial values. Rendering is unchanged: `none` resolves like the old unbounded default and `normal` computes to `0`, but both round-trip through `to_css`.
- `FontWeight::Absolute` holds an `f32` instead of a parley weight; `FontSynthesic` is renamed `FontSynthesisMode`; the stylesheet-parse and registered-font types are `#[non_exhaustive]`.
- `ImagesInput` had two incompatible public definitions — keep the base in `@takumi-rs/helpers`, re-export it from `takumi-js`, and rename the fetch-aware superset to `ManagedImagesInput`.

## Summary by CodeRabbit

* **New Features**
  * Added support for typed `gap` and `max-width`/`max-height` values, improving CSS round-tripping and layout handling.
  * Introduced a managed image input option for rendering, making remote image handling more consistent.
  * Exposed an optional `unstable` feature for advanced rendering/layout capabilities.

* **Bug Fixes**
  * Improved font-related styling and parsing consistency, including font synthesis, variations, and feature settings.
  * Updated error handling and selector parsing for more reliable stylesheet parsing.

### DIFF
--- a/.tegami/imagesinput-single-definition.md
+++ b/.tegami/imagesinput-single-definition.md
@@ -1,0 +1,13 @@
+---
+packages:
+  npm:takumi-js:
+    type: major
+---
+
+### Give `ImagesInput` one definition
+
+`takumi-js` and `@takumi-rs/helpers` each exported a different type named
+`ImagesInput`. The base shape now lives only in `@takumi-rs/helpers` and
+`takumi-js` re-exports it, so `ImagesInput` means the same thing everywhere. The
+fetch-aware superset that `render`/`renderSvg`/`renderAnimation` accept is now
+`ManagedImagesInput`.
--- a/.tegami/max-size-gap-initial-values.md
+++ b/.tegami/max-size-gap-initial-values.md
@@ -1,0 +1,15 @@
+---
+packages:
+  cargo:takumi-core:
+    type: minor
+  cargo:takumi:
+    type: major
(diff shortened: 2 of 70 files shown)


kane50613:
Superseding this approach. Sealing foreign types via `unstable`-gating treats the symptom; the root fix is encapsulating the core→backend boundary behind a core-owned type. Re-splitting into focused PRs: max-size/gap initial-values feature, the JS-surface fix, and the boundary encapsulation.

PR #927, opened 2026-07-07
Crate: fix decode/gradient/shadow panics and buffer-pool UB

kane50613:
## Why

Four render paths crash, hit undefined behavior, or over-allocate on hostile or degenerate input reachable from user CSS and author-influenced image URLs.

## What

- **Buffer pool bucket invariant**: an externally sized buffer released into the pool could later be `set_len` past its allocation (heap OOB write, UB). Release now buckets by the floor power of two the capacity guarantees, and `acquire_dirty` reserves before `set_len`.
- **Gradient stop snapping**: a gradient with more color stops than the LUT can hold panicked (`clamp` with `min > max`, then out-of-bounds indexing). The clamp is order-safe and the normalization passes stay within LUT bounds.
- **Shadow and blur sizing**: huge blur radii overflowed the `u32` shadow-buffer area, panicking or attempting a multi-GB allocation. Sizing uses saturating math and skips shadows above a 256 Mi-pixel budget.
- **Image decode limits**: decoders reject images beyond 8192x8192 (via `image::Limits` for PNG/JPEG, dimension checks for WebP) and GIFs beyond a total-frame pixel budget, stopping decode-bomb OOM. Documented in the images page.

Each fix ships with unit tests. Rendering output is unchanged for valid input.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added pixel-budget protections to reject oversized images and decode-bomb GIFs, and capped decode dimensions to 8192×8192.
  * Fixed gradient color-stop LUT handling to prevent panics when stop counts exceed LUT capacity.
  * Improved blur and shadow/drop-shadow rendering by using overflow-safe sizing and rejecting unsafe/extreme shadow allocations.
  * Updated buffer pool behavior so released/dirty buffers use consistent power-of-two bucketing and aren’t grown beyond their allocation.
* **Documentation**
  * Documented the new image decode limits, including the GIF frame budget.

### DIFF
--- a/.tegami/buffer-pool-capacity-invariant.md
+++ b/.tegami/buffer-pool-capacity-invariant.md
@@ -1,0 +1,13 @@
+---
+packages:
+  "cargo:takumi-raster": patch
+  "cargo:takumi": patch
+  "npm:@takumi-rs/core": patch
+  "npm:@takumi-rs/wasm": patch
+---
+
+### Fix buffer pool bucket capacity invariant
+
+Release now buckets a buffer by the floor power of two its capacity guarantees,
+and `acquire_dirty` reserves before `set_len`. A pooled buffer can no longer be
+lengthened past its allocation.
--- a/.tegami/gradient-lut-stop-clamp.md
+++ b/.tegami/gradient-lut-stop-clamp.md
@@ -1,0 +1,12 @@
+---
+packages:
+  "cargo:takumi-core": patch
+  "cargo:takumi": patch
+  "npm:@takumi-rs/core": patch
+  "npm:@takumi-rs/wasm": patch
+---
+
+### Fix gradient stop snapping panic for oversized stop lists
+
+Gradients with more color stops than the LUT can hold no longer panic. The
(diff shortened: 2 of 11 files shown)


kane50613:
Review feedback addressed in 981e4439 and 359ed586:

- Native WebP now reads dimensions with `WebPGetInfo` and checks the pixel budget before `WebPDecodeRGBA`.
- GIF decoding applies the same `Limits` as PNG/JPEG via `set_limits` before iterating frames.
- `release_bucket_index` guards `capacity == 0`.
- Dropped the redundant `.min(u32::MAX as f32)` clamp (float-to-int `as` casts saturate).

Not taken:

- Test for the `acquire_dirty` reserve path: reaching it requires a collision in the clamped top bucket (capacity >= 2^31), so a unit test would allocate multiple GiB. The guard is a defensive invariant.
- Shared `resolve_shadow_dimensions` helper: the two sites derive padding differently (f32 blur extent vs u32 padding), so the extraction would obscure more than it saves.

PR #929, opened 2026-07-07
Raster: reuse per-scene state across animation frames

kane50613:
## Why
Each animation frame re-ran work that is invariant across frames of the same
scene: `render_sequence_at_time` deep-cloned the whole `RenderOptions` tree per
frame, and every frame rebuilt the font snapshot (cloning the font collection
and repopulating per-script fallback tables). A 30 fps timeline paid both costs
30 times per second for output that never changed.

## What
Add a private `PreparedScene` that snapshots each scene's fonts once and wraps
its images and stylesheet in shared `Rc` handles. The per-frame render now clones
only the node and bumps those refcounts instead of deep-cloning options and
re-snapshotting. Both animation entry points route through it: `render_animation`
(eager) and `write_animation` (streaming). The single-image `render` path keeps
snapshotting per call. Frame output is bit-identical; the 220 regenerated golden
`.html` files were macOS regen noise with zero `.webp`/`.svg` diffs.

Bench validation deferred to CI and reviewer per the no-heavy-local-builds
constraint.

## Summary by CodeRabbit

  * Animation rendering now reuses per-scene state across frames, reducing repeated work during sequential playback.
  * Frame generation now uses precomputed scene data for more efficient rendering.

* **Bug Fixes**
  * Animation output remains unchanged while improving how frame-by-frame scene data is handled.

### DIFF
--- a/.tegami/reuse-animation-frame-state.md
+++ b/.tegami/reuse-animation-frame-state.md
@@ -1,0 +1,13 @@
+---
+packages:
+  "cargo:takumi-raster": patch
+  "cargo:takumi": patch
+  "npm:@takumi-rs/core": patch
+  "npm:@takumi-rs/wasm": patch
+---
+
+### Reuse per-scene state across animation frames
+
+Compute each scene's font snapshot once and share its image and stylesheet
+handles across frames instead of re-snapshotting and deep-cloning the whole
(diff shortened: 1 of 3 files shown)


kane50613:
@coderabbitai review

PR #931, opened 2026-07-07
JS: share the Renderer facade between the napi and wasm bindings

kane50613:
## Why
The `@takumi-rs/core` (napi) and `@takumi-rs/wasm` `Renderer` wrappers held byte-identical option types and four near-identical async methods, so every option had to be written twice. They had already drifted on cancellation: napi forwarded `signal` to the native call but never guarded the network-bound resource resolution, while wasm guarded around resolution but could not forward. Same public API, different abort behavior per backend.

## What
Add `prepareRenderInput` and the `BackendRenderOptions`/`BackendAnimationOptions`/`BackendSvgOptions` types to `@takumi-rs/helpers/renderer`. Both bindings now derive their option types from these and route every render method through the shared body, which checks `signal` before and after resolving fonts and images and returns it for the backend to forward when it can (napi does, wasm has no native signal). On native, an already-aborted signal now throws before any resource fetch. wasm keeps its `renderAsDataUrl` and `free`. New facade and abort-regression tests cover the shared body.

## Summary by CodeRabbit

* **New Features**
  * Shared rendering input handling now powers both native and web builds, keeping behavior aligned across backends.
  * Rendering now supports more consistent cancellation checks during resource loading.

* **Bug Fixes**
  * Improved abort handling so cancelled renders stop earlier and avoid unnecessary resource fetching.
  * Standardized font, image, and scene resource resolution across render, image, SVG, and animation flows.

* **Refactor**
  * Consolidated rendering option handling and input preparation into a single shared path.
  * Simplified backend-specific rendering methods by reusing common prepared inputs.

### DIFF
--- a/.tegami/share-renderer-facade.md
+++ b/.tegami/share-renderer-facade.md
@@ -1,0 +1,17 @@
+---
+packages:
+  npm:@takumi-rs/helpers:
+    type: minor
+  npm:@takumi-rs/core:
+    type: minor
+  npm:@takumi-rs/wasm:
+    type: minor
+---
+
+### Share the renderer facade between the napi and wasm bindings
+
+The napi and wasm `Renderer` wrappers now build on a shared `prepareRenderInput`
+helper in `@takumi-rs/helpers/renderer`, so their option types and render bodies
+live in one place. Both backends check `signal` before and after resolving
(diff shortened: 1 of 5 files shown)


kane50613:
@coderabbitai review

PR #964, opened 2026-07-14
Core: claim generic font families from the JS font API

kane50613:
## Why

The engine resolves CSS generic families (`monospace`, `sans-serif`, ...) through fontique's generic registry, and Tailwind's `font-mono` maps straight to the `monospace` generic — but the JS font API had no way to claim one. Registering a mono font did nothing for `font-mono`; every use needed an explicit `fontFamily`, and fonts registered by one component leaked into the default stack of the next render on a shared renderer with no way to scope them.

## What

- Font descriptors accept `generic` (CSS keyword: `monospace`, `serif`, `ui-monospace`, ...) across napi, wasm, and the shared helpers types; the value parses via `GenericFamily: FromStr` (delegating to parlance's keyword parser) and registers through `FontResource::generic_family`.
- Unknown keywords reject at registration with `FontError::UnknownGenericFamily`.
- napi test pins the behavior: a font registered with `generic: "monospace"` renders `font-family: monospace` byte-identical to naming the family, and without the claim it doesn't.
- Docs cover the field; the text-fit example loads Geist Mono with `generic: "monospace"` and keeps `font-mono`, and the example harness gets a fresh renderer per component so one component's fonts no longer leak into the next one's default stack.

## Summary by CodeRabbit

* **New Features**
  * You can now register fonts using CSS generic family keywords (e.g., `monospace`, `serif`, `system-ui`).
  * Font stacks can resolve to previously registered fonts via the `generic` setting (without an explicit family name).
  * Added support for generic fonts across JavaScript, Node, and WebAssembly environments.
* **Documentation**
  * Updated typography/font configuration docs and examples to include the new `generic` option.
* **Bug Fixes**
  * Font registrations are isolated between renders so configurations don’t persist unexpectedly.
* **Tests**
  * Added coverage for correct generic resolution and rejection of invalid generic keywords.


kane50613:
Also swaps the embedded last-resort Geist to wght 300..800: trimming the subset to `ccmp`/`kern`/`liga` (the `'*'` feature set carried ss01-10/aalt alternates a last-resort face never uses) pays for the extra weight range — 23,616 bytes vs 23,808 before. Verified weights 300/400/600/800 plus ligatures and kerning render correctly; 300 was previously clamped to 400.

PR #968, opened 2026-07-14
Helpers: support generic on googleFonts families

kane50613:
## Why

#964 lets a font claim a CSS generic family, but `googleFonts` had no way to pass one through — loading a mono family from Google Fonts still left `font-mono` unresolved.

## What

- `googleFonts` family objects accept `generic`; it propagates to every coverage subset the family produces and joins the subset's dedup key.
- `GenericFontFamily` moves to `fonts.ts` (renderer.ts already imports from it; the reverse import was circular).
- Docs list the field, and the helpers test pins propagation and the unclaimed default.

## Summary by CodeRabbit

* **New Features**
  * Google Fonts families can now specify a CSS generic family, such as `serif` or `sans-serif`.
  * The selected generic family is applied consistently across all loaded font subsets, improving font stack resolution.

* **Documentation**
  * Updated typography documentation with details about the new `generic` option.
  * Added a changelog entry describing generic family propagation.

### DIFF
--- a/.tegami/google-fonts-generic.md
+++ b/.tegami/google-fonts-generic.md
@@ -5,4 +5,4 @@ packages:
 
 ### Support generic on googleFonts families
 
-A family's `generic` (e.g. `"monospace"`) propagates to every loaded coverage subset, so generic stacks like `font-mono` resolve to it. Families Google Fonts categorizes as Monospace claim it automatically.
+A family's `generic` (e.g. `"monospace"`) propagates to every loaded coverage subset, so generic stacks like `font-mono` resolve to it.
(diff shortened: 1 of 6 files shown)


kane50613:
Follow-up: families Google Fonts categorizes as Monospace now claim `generic: "monospace"` automatically (explicit `generic` still wins). The catalog generator unions the metadata category with a word-bounded `mono`/`code` name check — Google's own category misses real mono faces (Cascadia Code/Mono, Noto Sans Mono, SUSE Mono, Atkinson Hyperlegible Mono, M PLUS Code Latin, Rubik Mono One), and the name rule recovers exactly those with zero false positives against the current 1942-family catalog. 58 names, ~1KB in the bundle, emitted at build time so it's auditable and tracks catalog updates.

kane50613:
Reverted the auto-monospace list — the heuristic surface (Google's own misclassifications + name rules) carries more edge cases than the feature is worth. `generic` stays explicit opt-in.

PR #979, opened 2026-07-16
Core: feed filter layers to resvg without a PNG roundtrip

kane50613:
## Why

`apply_svg_filter` unpremultiplied the layer, encoded it to PNG, and had resvg decode and re-premultiply it, only because `usvg::ImageKind` had no raw-pixel variant. Vendoring resvg (#978) lets us add one.

## What

- Add `ImageKind::Raw { width, height, data }` to the vendored usvg tree: premultiplied RGBA8, sized directly, decoded by wrapping the buffer in a pixmap.
- `apply_svg_filter` now hands the layer pixels to the resolver as `Raw`; the unpremultiply + PNG encode/decode path is gone.
- `showcase_chrome_text.webp` may shift within test tolerance: the filter input no longer goes through an 8-bit unpremultiply/premultiply quantization. Will patch the golden from the CI artifact if it drifts there.

Stacked on #978.

https://claude.ai/code/session_01VkYpKpgHYAd971sQRiJZDV


kane50613:
Tried the more direct route (skip `resvg::render`, call `filter::apply` straight on the layer pixmap): `showcase_chrome_text` panics at `lighting.rs:185` (`src.width == dest.width`). The default filter region is -10%/120% of the element bbox, so it extends past the canvas; `render_group` sizes the intermediate layer to bbox + region and crops back after applying, and that logic lives in the render path. Calling `filter::apply` directly would mean duplicating that layer-bounds handling here, so the render call stays and only the PNG roundtrip goes.

PR #1035, opened 2026-07-25
Raster: stop redoing two pieces of per-node and per-glyph work

kane50613:
## Why

Two pieces of work were done on every render and thrown away.

Every node deep-cloned its parent's `ComputedStyle` — a macro-generated struct covering every CSS longhand, plus two `HashMap`s — so that a loop over `@property` rules could conditionally mutate it. When no stylesheet declares an `@property` rule, and most do not, the loop body never runs and the function returns an exact copy of its input.

Underlined text rasterized each glyph outline twice. The paint goes through the process-wide glyph-mask cache; the skip-ink bounds went through a direct `render_mask` call that neither reads nor fills that cache, so it repeated in full for every occurrence of the glyph in every underlined run, and again on every later render of the same text. `text-decoration-skip-ink: auto` is the CSS default, so this was the ordinary path for links and headings rather than an edge case.

## What

`registered_custom_property_parent_style` returns a `Cow` and borrows when no sheet declares an `@property` rule. The guard keys on `property_rules().is_empty()` rather than on the media-query match, so it is trivially equivalent to the old behavior: a sheet whose rules all fail to match still takes the clone.

Skip-ink now calls `get_or_render_cached_mask` at subpixel bucket 0, in the same key space the paint path uses. `bucket_zero_mask_matches_untransformed_render` pins the assumption that makes this safe: bucket 0 applies an identity translation, and tiny-skia returns the path unchanged for that, so the mask is byte-identical to `render_mask(paths, None, None)`.

Worth being precise about the win, since sharing the key space does not always mean sharing the entry. The paint path picks its bucket from the glyph's fractional x position, so text laid out at fractional advances paints from buckets 1-3 and skip-ink's bucket-0 entry serves only skip-ink. What the change buys in that case is that the bounds survive across runs and across renders instead of being rasterized from scratch every time; where the paint does land on bucket 0, the two share one rasterization outright.

Rendered output is unchanged: the full fixture suite produced zero `.svg` and zero `.webp` diffs.

Neither win is measured. The redundancy is certain from the call graph, the magnitude is not — the benchmark question is worth answering separately.

### DIFF
--- a/.tegami/render-path-redundancies.md
+++ b/.tegami/render-path-redundancies.md
@@ -1,0 +1,13 @@
+---
+packages:
+  "cargo:takumi-core": patch
+  "cargo:takumi-raster": patch
+---
+
+### Stop redoing two pieces of per-node and per-glyph work
+
+Every node deep-cloned its parent's `ComputedStyle`, a struct covering every CSS longhand plus two `HashMap`s, so that a loop over `@property` rules could conditionally mutate it. Stylesheets that declare no `@property` rule, which is most of them, paid the clone and got an exact copy back. The function returns a `Cow` now and borrows in that case.
+
+Underlined text rasterized each glyph outline twice: once through the shared glyph-mask cache for the actual paint, and once through a direct uncached call that only measured skip-ink bounds. The second pass neither read nor filled the cache, so it repeated for every occurrence of the glyph. It now goes through the same cache at subpixel bucket 0, which a new test pins as identical to the untransformed rasterization. `text-decoration-skip-ink: auto` is the CSS default, so this was the ordinary path for links and headings.
+
+Rendered output is unchanged.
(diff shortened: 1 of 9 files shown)


kane50613:
On the `cache_signature() << 2` aliasing note: skipping, with reasoning.

The shift is not introduced here. `glyph_cache_key_and_offset` in `text_drawing.rs` has built the paint path's key as `(glyph_signature << 2) | bucket_x` since the mask cache landed, so the key space has always been the top 62 bits of the signature plus a 2-bit bucket. This PR deliberately reuses that packing rather than inventing a second one: bucket 0 is the same entry the paint path stores, which is the whole point of the change, since the two consumers now share one rasterization instead of doing it twice.

Narrowing an xxh3 digest from 64 to 62 bits leaves collision probability around 2^-31 at a billion distinct outlines in a single process. If that were the real risk, the paint path would already be wrong.

Centralizing the packing into one helper is fair as tidying, but it belongs with the paint path rather than in a change that only adds a second caller, so I have left it out here.

kane50613:
@coderabbitai review

kane50613:
Folded into #1036, which now carries all four commits.

PR #1036, opened 2026-07-25
Core: cut redundant per-node and per-glyph work, and put glyph caches in one pool

kane50613:
## Why

Work on the glyph path was being redone or mis-budgeted in four places. Every node deep-cloned its parent's `ComputedStyle` so a loop over `@property` rules could mutate it, though most stylesheets declare no such rule. Underlined text rasterized each outline twice, once for the paint and once uncached for skip-ink bounds. Text stroke and faux-bold never touched the cache at all, so CJK bold — synthesis triggers at weight 600 and up — re-rasterized a stroke per glyph per render. And the glyph budget was split exactly in half between resolved outlines and rasterized masks, a ratio the content sets rather than the code: a 24px CJK outline is ~3.4 KB against a 576-byte mask, so the resolved half thrashes at ~1200 glyphs while the mask half never fills, and a 96px heading inverts it.

## What

The style helper returns a `Cow` and borrows when no `@property` rule exists. Skip-ink, stroke and faux-bold all go through the mask cache, keyed on the stroke as well as the outline. The two glyph caches become one pool under one budget, keyed by `(u64, kind)` so neither stage loses key bits, and cached resolved glyphs are handed out as `Arc` so a hit stops copying outline commands. `Placement` moves to `takumi_core::geometry` and replaces `ResolvedGlyphPlacement`.

Five goldens move: stroked masks now land on the same quarter-pixel grid the fill already used, which shifts antialiasing by a fraction of a pixel and, along the way, stops the stroke drifting from the fill it outlines. The renders are visually identical — I compared them. Everything else is byte-identical.

## Summary by CodeRabbit

* **Performance**
  * Improved text/glyph rendering speed using a process-wide cache for resolved glyph outlines and rasterized glyph masks.
  * Reduced copying by sharing resolved glyph data and using reference-counted glyphs across layout and raster pipelines.
* **Reliability**
  * Prevented cache collisions between outline and mask entries and ensured concurrent misses coalesce to a single in-flight result.
  * Failed glyph resolutions no longer pollute the cache; stroked/faux-bold mask rendering stays aligned to fill antialiasing.
* **Documentation**
  * Updated glyph-cache notes/version bump guidance and documented the new pixel-space `Placement` API.


kane50613:
@coderabbitai review

kane50613:
Both looked at, one fixed.

The packed key was a real defect: `key << 1 | kind` drops bit 63, so keys 0 and `1 << 63` aliased within the same kind. The cache is keyed by `(u64, GlyphKind)` now, so both stages keep all 64 bits and the two can never collide. `the_whole_key_is_kept` covers exactly that pair.

`ENTRY_OVERHEAD` on resolved entries is skipped: `ResolvedGlyph::estimated_bytes` already adds its own 64 in every arm, so adding it again at the cache would double-count. Masks add it there because `Vec::capacity` carries none.

PR #1037, opened 2026-07-25
Helpers: keep the glyph budget setter off the backend import graph

kane50613:
Fixes the packaging check, which has been red on `master` since #1033.

## Why

`takumi-js` re-exported `setGlyphCacheMaxBytes` from `./import`, the module that resolves the backend. That pulled `#backend` into the type graph of the root entry, and `attw` reports it as an internal resolution error for the root export:

```
"takumi-js"   node16 🟢   bundler 🟢   node10 (ignored) 🥴 Internal resolution error
```

Every other entry only has the ordinary node10 `💀 Resolution failed`, which the `node16` profile ignores. The internal-resolution one still fails the run, so `attw --pack .` exits 1 and `publish-lint` never reaches `publint`.

Making the parameter list narrower was not enough. The cause is the re-export itself: nothing in `render`'s public types names anything from `./import`, so `#backend` gets shaken out of the declarations, and re-exporting a value from that module puts it back.

## What

The setter and its recorded value move to `src/glyph-cache.ts`, a leaf that imports nothing. `getImports` hands the recorded budget to the backend as it loads, through a structurally typed applier, so the leaf never has to name the backend type.

The public function also stops being async. It no longer has to resolve anything to record a number, and applying the value during backend load removes the ordering hazard the async version had: a caller who awaited it on a runtime that needs an explicit WASM `module` would have resolved the backend without one.

Verified locally with the same command CI runs: `attw --pack . && publint --strict .` now exits 0, `tsc --noEmit` is clean, and the 20 `takumi-js` tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added synchronous glyph cache budget configuration in `takumi-js`.
  * Ensures the configured cache limit is applied automatically when the rendering backend finishes loading (avoids timing issues).

* **Documentation**
  * Updated performance and API documentation examples to remove `await` and reflect the new synchronous behavior.

### DIFF
--- a/.tegami/set-glyph-cache-max-bytes.md
+++ b/.tegami/set-glyph-cache-max-bytes.md
@@ -13,4 +13,4 @@ The resolved-glyph and glyph-mask caches share an 8 MiB budget that no binding e
 
 The default suits Latin text. A CJK outline runs a few kilobytes, so 8 MiB holds on the order of a thousand of them and a page of Chinese re-rasterizes glyphs it evicted a moment earlier.
 
-`takumi-js` forwards it too. That one is async, since it has to resolve the backend before it can set anything.
+`takumi-js` forwards it too. That one records the budget and hands it to the backend as it loads, so it stays synchronous and cannot race the resolution.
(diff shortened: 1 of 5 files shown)


kane50613:
@coderabbitai review

kane50613:
Fixed in bffa838. You were right that the contract and the code disagreed: recording the value only helped a backend that had not loaded yet, and nothing forwarded a later call.

The module now keeps the loaded backend and forwards to it directly, so the setter works whenever it is called. What remains true, and is what the doc now says, is that the backend itself reads the budget when a cache is first used, so a call after the first render does not resize a cache already in place. That limit belongs to the Rust side rather than to this forwarding layer.

PR #1042, opened 2026-07-27
Add new showcase entry for Rail Radar

R4ULtv:
## Summary by CodeRabbit

* **New Features**
  * Added the “Rail Radar” project to the showcase gallery, including its preview image and destination link (with set 1200×630 dimensions).

### DIFF
--- a/docs/app/data/showcase.ts
+++ b/docs/app/data/showcase.ts
@@ -130,7 +130,7 @@ export const showcaseProjects: Project[] = [
   {
     title: "Rail Radar",
     image: "https://www.railradar24.com/og?id=CH03000",
-    url: "https://railradar.com",
+    url: "https://www.railradar24.com",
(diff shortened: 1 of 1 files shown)


kane50613:
The domain is not reachable (403), and URL points to a different website

R4ULtv:
Sorry, I misspelled the URL and used the wrong domain.

R4ULtv:
The URL has been fixed, but I completely forgot that the service is currently restricted to users in the EU and the US through the Vercel Firewall. That’s why requests from other regions return a 403, and also why I didn’t see any corresponding 403s in the application logs.

I don’t want to remove those geographic restrictions right now, since the service is primarily intended for European users, so I’m going to close this PR for the time being.

I’m really sorry for overlooking this and for wasting your time.

PR #1052, opened 2026-07-29
JS: add `ImageResponse.render` for a fully rendered response

### LINKED ISSUE #1050: ETag header
### Pre-submission Checklist

- [x] I have searched [existing issues](https://github.com/kane50613/takumi/issues) and this feature has not been requested

### Problem Statement

The [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag) header can be used to improve the efficiency of caching while also preventing stale entries. 

Certain images like favicons or some OG images may rarely change, but setting aggressive cache headers runs the risk of showing stale content whenever it does change. Hence, setting an ETag header would be ideal.

There isn't an easy way to set an ETag on the consumer's side of Takumi (the easiest I think would be to await the ImageResponse, creating a hash of the bytes, and then recreating it, which is very inefficient) and I believe the image gets [fully loaded into memory in Takumi's side](https://github.com/kane50613/takumi/blob/357c5fdd2440f4f2ac49d7ac5b2edf6559c9fb24/takumi-js/src/response/index.ts#L43).

### Proposed Solution

[globalThis.crypto.subtle.digest](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) is available in [Node.js](https://nodejs.org/api/webcrypto.html#cryptosubtle) (since 15), [browsers](https://caniuse.com/mdn-api_subtlecrypto_digest) (Baseline Jan 2020), and [Cloudflare Workers](https://developers.cloudflare.com/workers/runtime-apis/web-crypto/) (unflagged).

[Uint8Array.toHex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toHex) doesn't have very good support but could be replaced with a simple polyfill.

### Proposed API / Usage

Either as an image option (`etag?: boolean`) or even at all times (SHA256 for an image should be pretty fast on all systems) would probably be ideal

### Alternatives Considered

_No response_

### Related Package

@takumi-rs/image-response

### Additional Context

_No response_

kane50613:
## Why

Closes #1050. `new ImageResponse(...)` keeps its `next/og` shape, which means the framework reads its headers while the render is still in flight, so no `ETag` can reach them. Callers who wanted one had to await the response, hash the bytes, and rebuild it. The body is also a `ReadableStream` over a single chunk that is already fully in memory, so the runtime pumps a stream and falls back to chunked encoding for bytes whose length it could have known.

## What

`ImageResponse.render(element, options)` takes the same options and returns `Promise<Response>`. It awaits the render, then builds the response from the finished bytes: no stream, known body length, and a strong `ETag` that is their SHA-256 digest. An `etag` passed through `headers` wins. The constructor is untouched.

Takumi never sees the request, so answering `If-None-Match` with a 304 stays the server's or CDN's job; the `ETag` is what makes that possible.

The digest uses global Web Crypto, which Node exposes from 19 onwards. `engines` stays at `>=18` rather than dropping a runtime in a minor, and a `node:crypto` fallback is deliberately not there: an unconditional `node:` import is the same trap that broke bundler resolution in #952.

Two fixes fell out of the shared helpers: a rejecting `onError` no longer replaces the render error (or escapes as an unhandled rejection from the stream's `start`), and the one `as ArrayBufferView<ArrayBuffer>` cast is now a narrowing helper.

## Summary by CodeRabbit

* **New Features**
  * Added `ImageResponse.render` for producing fully rendered image responses with a strong `ETag`.
  * Automatically sets the image content type and preserves caller-provided `ETag` values.
  * Rendering failures now reject the returned promise while invoking the configured error handler.

* **Bug Fixes**
  * Improved streaming response error propagation and output byte handling.

* **Documentation**
  * Documented rendering, caching behavior, `ETag` generation, and runtime requirements.

### DIFF
--- a/.tegami/image-response-buffered.md
+++ b/.tegami/image-response-buffered.md
@@ -1,0 +1,13 @@
+---
+packages:
+  npm:takumi-js:
+    type: minor
+---
+
+### Add `ImageResponse.buffered` for a fully rendered response
+
+`ImageResponse.buffered(element, options)` awaits the render and returns a
+`Response` carrying the finished bytes instead of a stream, so the runtime knows
+the body length and the headers carry a strong `ETag` of the image. The
+constructor cannot set one, since its headers are read while the render is still
+in flight.
--- a/.tegami/image-response-render.md
+++ b/.tegami/image-response-render.md
@@ -4,9 +4,9 @@ packages:
     type: minor
 ---
 
-### Add `ImageResponse.buffered` for a fully rendered response
+### Add `ImageResponse.render` for a fully rendered response
 
-`ImageResponse.buffered(element, options)` awaits the render and returns a
+`ImageResponse.render(element, options)` awaits the render and returns a
 `Response` carrying the finished bytes instead of a stream, so the runtime knows
 the body length and the headers carry a strong `ETag` of the image. The
 constructor cannot set one, since its headers are read while the render is still
--- a/docs/content/docs/image-response.mdx
+++ b/docs/content/docs/image-response.mdx
@@ -76,15 +76,15 @@ new ImageResponse(<OgImage />, {
 });
 ```
 
-## Buffered responses and ETag
+## Awaiting the render, and ETag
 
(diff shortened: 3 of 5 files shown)


kane50613:
I'll close this for now since hashing from input data or the tree + options could be a better option than hashing the entire image buffer

PR #1113, opened 2026-08-06
PDF: tag decorations as artifacts and split the crate into modules

kane50613:
## Why

Backgrounds and borders were painted outside any tagged content sequence. veraPDF reported untagged content on every page that drew one, so no document could pass PDF/UA-1. `takumi-pdf/src/lib.rs` had also grown to 2934 lines, `takumi-pdf-wasm/src/lib.rs` to 690.

## What

Each of `emit_background`, `emit_background_layers`, and `emit_borders` now opens its own artifact sequence after its early returns. A decoration is tagged where it is painted, so a new decoration kind cannot be added without one.

The two `lib.rs` files split into modules along their existing seams. The move changes no behavior: 458 lines are left in the pdf crate, 222 in the wasm crate.

`example/e-invoice` renders an invoice with a Factur-X payload attached, and its README documents validating the output with veraPDF and Mustang. The output passes veraPDF for both `3b` and `ua1`. A conforming Factur-X container also needs an `fx:` XMP block, which the README calls out; #1114 adds the API for it.

The 16 changed PDF fixtures carry the new artifact markers.

## Summary by CodeRabbit

- **New Features**
  - Added an executable e-invoice example that generates tagged PDF/A-3B invoices with embedded Factur-X XML.
  - Added support for PDF metadata, attachments, tagging, archival standards, headers, footers, page numbering, hyperlinks, and document outlines.
  - Added flexible page sizes, margins, viewport rendering, image inputs, font registration, gradients, borders, SVGs, and decorative content.
  - Added validation for dates and rendering options, with clearer handling of invalid formats and conflicting settings.

- **Documentation**
  - Added setup, usage, accessibility, and validation guidance for the e-invoice example.


kane50613:
@coderabbitai full review

PR #1114, opened 2026-08-06
PDF: write custom XMP schemas

kane50613:
## Why

An attached invoice XML becomes a Factur-X file only once the XMP packet says so. Other payload-specific metadata works the same way, from C2PA claims to retention policies. A general PDF engine does not define those schemas, and the API had no way to supply them.

## What

`metadata.xmp` takes namespaces as data: a prefix, a URI, and the properties written under it. The renderer writes each property as a value and describes it in the `pdfaExtension:schemas` entry PDF/A requires, so the two cannot disagree, and both go through the packet writer rather than into the serialized bytes.

`example/e-invoice` now emits a Factur-X MINIMUM invoice: veraPDF passes it as PDF/A-3B and PDF/UA-1, and Mustang reports the whole file valid.

CI installs veraPDF and validates every generated fixture against the standard it claims, so a conformance regression fails the build instead of waiting for someone to run a validator by hand. The remaining crate-local dependency versions also move to `[workspace.dependencies]`, where `takumi-pdf` was the last holdout.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom XMP metadata schemas and properties in generated PDFs.
  * Added Factur-X metadata support to the e-invoice example, producing PDF/A-3B and PDF/UA-1 documents.
  * Custom metadata is included in PDF/A extension schema descriptions.
* **Bug Fixes**
  * Improved handling of backgrounds and borders as tagged-content artifacts.
  * Added validation and clear errors for invalid XMP schema values.
* **Documentation**
  * Updated e-invoice and PDF metadata documentation with validation and metadata requirements.
* **Tests**
  * Added automated veraPDF validation for PDF/A and PDF/UA fixtures.


kane50613:
@coderabbitai review

kane50613:
The Rust `PdfMetadata` struct is not a published API: `takumi-pdf` sets `publish = false`, so no external crate can construct it. The published surface is the npm package, where `xmp` is optional. Pinned the veraPDF download by checksum in the same commit.

kane50613:
@coderabbitai full review

PR #1116, opened 2026-08-06
PDF: link to anchors inside the document

kane50613:
## Why

`href` values were kept only for `http`, `https`, `mailto` and `tel`, so a table of contents rendered as plain text. The pieces to resolve a target were already there: headings become outline destinations the same way.

## What

Anchors now record every element that carries an `id`, and an `href="#id"` becomes a link annotation pointing at the page holding that element. A fragment matching no element is dropped, so a reader never gets a clickable box that goes nowhere.

`Node::id` becomes public in takumi-core, alongside the existing `href`, `alt` and `tag_name` accessors.

The `report-links-outline` fixture gains a link into the appendix, which sits behind a forced page break, and a fragment pointing at nothing.

## Summary by CodeRabbit

* **New Features**
  * Fragment links now resolve by element `id` (percent-decoded) in rendered PDFs.
  * Link targets and document outline destinations are handled consistently across paginated and single-page output.
  * Public access to element IDs is now available.

* **Bug Fixes**
  * Links to nonexistent fragments no longer generate invalid link annotations.
  * Duplicate element IDs keep the first resolved target.

* **Documentation**
  * Added/expanded guidance for external schemes (HTTP/HTTPS/mailto/tel) and internal fragment behavior, including missing targets.


kane50613:
@coderabbitai review

kane50613:
@coderabbitai full review

PR #1117, opened 2026-08-06
PDF: place background layers with size, position and repeat

kane50613:
## Why

Gradient layers stretched across the whole box no matter what `background-size`, `background-position` or `background-repeat` said. Those three are pure geometry in Blink too, and PDF has a native tiling pattern, so nothing about them needs a rasterizer.

## What

Each layer resolves to a tile size, a first-tile origin and a step. A layer that tiles becomes one PDF tiling pattern holding a single drawn tile, so a repeated gradient costs one shading rather than one per tile; `space` widens the step, `round` rescales the tile to fit a whole number of them.

The positioning area is still the border box, so `background-origin` is unread, and url() layers are still unsupported. Defaults resolve to the old behavior, which is why no existing fixture moved.

## Summary by CodeRabbit

* **New Features**
  * PDF backgrounds now support `background-size`, `background-position`, and `background-repeat`, including `no-repeat`, `repeat`, `space`, and `round`.
  * Repeating gradient backgrounds use reusable tiled patterns for improved rendering.
* **Bug Fixes**
  * Repeating radial gradients are capped at 512 cycles to prevent excessive expansion.
  * Improved pagination handling for content crossing page boundaries.
* **Documentation**
  * Updated PDF documentation and release notes to describe background placement, border-box positioning, and current limitations.


kane50613:
@coderabbitai review

kane50613:
@coderabbitai full review

PR #1118, opened 2026-08-06
PDF: clip elements with clip-path

kane50613:
## Why

`clip-path` was ignored. In Blink it is overwhelmingly closed-form geometry: basic shapes resolve to a path, and only clips it cannot flatten fall back to a raster mask. PDF has a native clipping path, so nothing here needs a rasterizer.

## What

`inset()`, `ellipse()`, `polygon()` and `path()` resolve to a path and clip the element before anything is painted, so decorations are clipped too. The shape's own `fill-rule` wins over `clip-rule`. A shape that resolves to no area clips everything away, which is what CSS asks for; a shape that cannot be resolved at all, which today means `path()` in a build without core's `svg` feature and its parser, leaves the element unclipped instead.

The geometry lives in `takumi_core::layout::clip`, ported from the raster backend's copy. Raster still runs its own copy for now: switching it over changes its output and churns the goldens, so it belongs in its own change.

The fixture renders one cell per shape and asserts the clip operators reach the content stream, non-zero and even-odd counted separately, which needs `flate2` as a dev-dependency to read the deflated streams.

## Summary by CodeRabbit

* **New Features**
  * Added PDF `clip-path` support for `inset()`, `ellipse()`, `polygon()`, and SVG `path()` shapes.
  * Added support for fill rules, including even-odd clipping.
  * Added handling for zero-area clips that hide content completely.

* **Documentation**
  * Updated feature documentation and release notes to reflect PDF clipping support.


kane50613:
@coderabbitai review

kane50613:
@coderabbitai full review

PR #1119, opened 2026-08-06
PDF: draw box-shadow

kane50613:
## Why

`box-shadow` was dropped entirely, which is the most visible gap for anything designed on the web. In Blink everything but the blur is rect arithmetic, and the blur is the only part that needs a rasterizer.

## What

Outer and inset shadows paint as vector fills. The shadow shape is the border box spread and offset, and the element's own shape is appended so an even-odd fill leaves a ring, which is how Blink keeps a shadow from painting under an opaque box.

The blur is approximated: eight bands walking inward from a full blur radius out, each band's opacity chosen so the fills composite to the coverage a Gaussian would leave at that distance. A shadow with `blur-radius: 0` is one exact fill and needs no approximation. Raising the band count or rasterizing just the shadow layer are both open to us later; the constant carries a `ponytail:` note saying so.

Shadows sit in their own artifact sequence, like the other decorations.

## Summary by CodeRabbit

* **New Features**
  * Added PDF support for CSS `box-shadow`, including outer and inset shadows.
  * Supports offsets, spread, rounded corners, transparent borders, and blurred shadows.
  * Blur effects are rendered with layered opacity bands for smooth visual results.

* **Tests**
  * Added coverage for sharp, blurred, inset, and translucent shadows in generated PDFs.


kane50613:
@coderabbitai review

kane50613:
@coderabbitai full review

PR #1120, opened 2026-08-06
PDF: apply the color filter primitives

kane50613:
## Why

`filter` was ignored. In Blink the primitives split cleanly: `grayscale`, `sepia`, `saturate`, `hue-rotate`, `invert`, `brightness` and `contrast` are color-matrix and transfer-function operations, while only `blur` and `drop-shadow` need a convolution. The first group is a linear transform of the source color, so a vector backend can fold it into the colors it writes.

## What

Those primitives compose into one matrix per stacking context, inherited by descendants, and apply to every color the subtree paints: fills, borders, text, decorations, gradient stops and decoded image pixels. `opacity()` folds into the alpha channel.

The equivalence has a limit worth knowing: transforming each color before compositing matches compositing first only while the filtered content is opaque. Overlapping translucent content differs, which is the price of not rasterizing the subtree.

`blur()` and `drop-shadow()` stay ignored, as do referenced SVG filters. The fixture paints one red square per filter and asserts each lands a different color; `invert(1)` comes out exactly one minus the source, and `grayscale(1)` exactly the Rec. 709 luma.

## Summary by CodeRabbit

* **New Features**
  * Added support for linear CSS color filters in PDF output, including brightness, contrast, grayscale, sepia, invert, opacity, saturation, and hue rotation.
  * Filters now apply consistently to backgrounds, gradients, borders, text, shadows, images, and nested stacking contexts.
  * SVG images are rasterized when color filters are applied.

* **Documentation**
  * Documented supported filters and limitations, including unsupported blur, drop-shadow, and referenced SVG filters.


kane50613:
@coderabbitai review

kane50613:
@coderabbitai full review

PR #1121, opened 2026-08-06
PDF: fade elements with mask-image

kane50613:
## Why

`mask-image` was ignored. Blink implements masks as a compositing operation rather than a path, which sounds like it forces a rasterizer, but PDF has native soft masks: a form XObject whose content supplies the alpha. A gradient mask can therefore stay vector end to end.

## What

An element with gradient mask layers draws them into their own stream and pushes it as an alpha soft mask, which then covers the element and its descendants. The mask content is the same gradient machinery the backgrounds use, so it costs a shading, not a bitmap.

`url()` mask sources are still ignored. The fixture masks a solid fill and a gradient fill, and asserts one alpha soft mask lands per element.

## Summary by CodeRabbit

* **New Features**
  * Added PDF support for CSS `mask-image` layers using linear, radial, and conic gradients.
  * Preserves vector content while applying alpha-based soft masks to elements and their descendants.
  * Supports mask sizing, positioning, and repeating controls, including tiled mask placement.
  * Maintains correct interaction with opacity filters.
* **Documentation**
  * Documented PDF mask support and clarified limitations, including unsupported URL-based mask and background layers.
* **Tests**
  * Added coverage verifying masked PDF output and generated alpha soft masks.


kane50613:
@coderabbitai review

kane50613:
@coderabbitai full review

PR #1125, opened 2026-08-06
PDF: rasterize blur and drop-shadow filters like Chromium layers

kane50613:
## Why

`filter: blur()` and `drop-shadow()` were ignored: they are convolutions, and a vector PDF has no operator for them.

## What

A filtered stacking context now rasterizes through a new `takumi_raster::rasterize_node` entry point, the way Chromium renders a filtered layer offscreen, and embeds as an image with the filter's bleed margin. All filter functions in the list apply during that pass, so ordering matches CSS; color-only filter lists keep the existing vector path. Ships behind the `raster-filters` feature, on by default, at about +8% wasm size. The raster pass runs at 1x for now: supersampling needs the subtree's sizing to carry a device-pixel ratio first.


kane50613:
Closing: the raster pass costs about 8% wasm size, which is not worth two filter functions. The approach (rasterize_node + Chromium-style offscreen layer) is recorded if demand ever justifies it.

PR #1128, opened 2026-08-06
Docs: add PDF use case and migration guides

kane50613:
## Why

The PDF section was a flat list of feature guides. Nothing told a reader whether takumi-pdf fits an invoice run, and nothing helped someone already on Puppeteer or react-pdf work out what changes.

## What

Two new sidebar categories. **Use cases**: invoices (Factur-X packaging, veraPDF and Mustang validation), reports (chapters, bookmarks, long tables, batch runs), tickets (single-page output). **Migration**: from Puppeteer and from react-pdf, each with an option map and an honest list of what does not carry over.

Along the way: icons on every PDF page, twoslash cuts no longer hide real code, two `[!code]` spans that covered the wrong lines, and the stale note claiming takumi cannot write the Factur-X XMP schema.

## Summary by CodeRabbit

* **Documentation**
  * Added migration guides for Puppeteer and React PDF, including API mappings, layout differences, and unsupported features.
  * Added comprehensive guides for invoices, reports, tickets, labels, and single-page PDFs.
  * Expanded coverage of PDF/A, Factur-X/ZUGFeRD, accessibility, pagination, headers, fonts, links, and outlines.
  * Documented image fetching, security controls, validation workflows, and reproducible PDF output.
  * Improved PDF documentation navigation with organized guides, use cases, and migration sections.
  * Corrected code-example highlighting and updated page icons and annotations.

### DIFF
--- a/docs/content/docs/index.mdx
+++ b/docs/content/docs/index.mdx
@@ -91,7 +91,7 @@ const png = await render(
   {
     width: 1200,
     height: 630,
-    // [!code highlight:3]
+    // [!code highlight:7]
     fonts: googleFonts([
       {
         name: "Fraunces",
--- a/docs/content/docs/load-images.mdx
+++ b/docs/content/docs/load-images.mdx
@@ -62,7 +62,7 @@ import { ImageResponse } from "takumi-js/response";
 
 export function GET() {
   return new ImageResponse(<OgImage />, {
-    // [!code ++]
+    // [!code ++:10]
     images: [
       {
         src: "my-logo",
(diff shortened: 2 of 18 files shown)


kane50613:
Applied four of the six review notes, plus the same import fix across the existing PDF pages:

- **from-puppeteer** no longer claims the option names match. Names and input types differ; the map below the paragraph is the source of truth.
- **from-react-pdf** applies the `96 / 72` factor to lengths only, and names the unitless properties that carry over as they are.
- **invoices** points the per-file date at `modificationDate`, with a line noting the `metadata.creationDate` fallback.
- **twoslash cuts** now hide only the `declare` scaffolding. Every snippet in the PDF section shows its imports, so it can be copied as-is.

Two skipped:

- **`comparison` under Migration.** It benchmarks takumi-pdf against exactly the two engines the migration pages cover, so it is the page a reader wants next when deciding whether to move. Grouping it there is deliberate.
- **Long tables.** The text already states the constraint: `break-inside-avoid` moves a group that no longer fits to the next page, and a group that outgrows the page splits again. Takumi has no per-page-break hook, so inserting a header after every break is not available to write instead.

PR #1131, opened 2026-08-06
PDF: pack the structure tree into an object stream

kane50613:
Stacked on #1130.

## Why

A tagged document writes one dictionary per structure element. On a two-page invoice that is 97 objects like

```
5 0 obj
<</Type/StructElem/S/P/P 102 0 R/K[0]/Pg 108 0 R>>
endobj
```

Each is its own uncompressed indirect object, and together they are a third of the file. They are also nearly identical, which is exactly what a compressed stream is good at.

## What

They move into a single object stream. `pdf-writer` writes only free and occupied cross-reference rows, so the type-2 rows the packed objects need are written afterwards, over the fixed-width rows of the cross-reference stream. The object stream is gated on PDF 1.5, the same version that brought the cross-reference stream pointing at it.

Reading a chunk back into its objects is a parse, and a text string inside a dictionary can spell `endobj`, so a candidate boundary is only accepted when the next object's header follows it. Anything that does not read back as the plain sequence of dictionaries expected leaves the chunk written as it stands.

## Numbers

| | before | after |
| --- | ---: | ---: |
| `invoice.pdf` | 34,390 | 23,615 (−31.3%) |
| `paged-lines.pdf` | 11,881 | 9,324 (−21.5%) |
| 41 golden fixtures | 415,367 | 352,200 (−15.2%) |
| 300 renders, median of 3 | 14.73 ms | 14.77 ms |

Both tables measure against master, so they include #1130.

## Checks

veraPDF passes PDF/A-2b, PDF/A-3b, PDF/A-4 and PDF/UA-1 on the fixtures, which means a strict parser reads the structure tree through the type-2 rows. Ghostscript renders every checked fixture byte-identically to master.

The two tests that searched the raw bytes for a structure element now search the inflated streams, since that is where those elements live.

## Summary by CodeRabbit

* **New Features**
  * Reduced the size of tagged PDF documents by compressing structure-tree data.
  * Preserved compatibility with PDF/A and PDF/UA output.
  * Maintained correct tagging and cross-reference behavior across supported PDF versions.

* **Tests**
  * Expanded PDF fixture validation to verify content within compressed streams and ensure accessibility-related features remain intact.


kane50613:
@coderabbitai review

kane50613:
All three applied.

The version check is the real one: `take()` ran before the `filter`, so a document below PDF 1.5 dropped the chunk entirely instead of falling through to the normal write. The check now gates the whole branch, and the restore-on-failure path is the only thing that puts the chunk back.

The offset field is also refused above eight bytes now, since that is the width it is read from, and the test helper's comment says what it does rather than why it used to be unnecessary.

No fixture for tagged PDF 1.4: nothing selects that version today, the archival levels name it only as their floor. The path is now the same one every other version takes when packing declines, which the existing fixtures already cover.

kane50613:
@coderabbitai review

PR #1132, opened 2026-08-06
PDF: write less per embedded font

kane50613:
Stacked on #1131.

## Why

Three things every embedded font was paying for, none of which a reader looks at.

`cid.default_width(0.0)` meant no glyph could ever match `/DW`, so `/W` had to spell out every width. A monospaced or CJK face writes the same number thousands of times.

`CIDSet` exists for PDF/A-1b. PDF/A-1 prohibits transparency, so takumi does not offer it, and every other level treats the stream as optional.

`/FontBBox` was computed by tracing the outline of every glyph in the subset, on every render, to tighten a rectangle whose only requirement is that it encloses them.

## What

`/DW` takes the width most glyphs share and `/W` lists the exceptions. The `CIDSet` stream and its reference are gone, along with the version predicate that guarded them. `/FontBBox` comes from the subset font's own box.

## Numbers

Against master, so these include #1130 and #1131:

| | before | after |
| --- | ---: | ---: |
| `invoice.pdf` | 34,390 | 23,492 (−31.7%) |
| `paged-lines.pdf` | 11,881 | 9,207 (−22.5%) |
| 41 golden fixtures | 415,367 | 348,678 (−16.1%) |

The dropped outline pass is within noise on these fixtures, whose subsets are small; it scales with the number of glyphs a document uses.

veraPDF passes PDF/A-2b, PDF/A-3b, PDF/A-4 and PDF/UA-1. Ghostscript renders every checked fixture byte-identically to master, which is the check that matters for `/DW`: the widths reaching the reader are the same numbers, said more briefly.

## Summary by CodeRabbit

* **Performance**
  * Reduced PDF font metadata and width data to produce more compact, efficient PDF files.
  * Improved font subsetting by using only the necessary glyph information.

* **Compatibility**
  * Updated generated PDF font structures to align with modern PDF formatting practices by removing unnecessary CID set data.

* **Documentation**
  * Added a changelog entry describing the reduced CID font data and related PDF output improvements.


kane50613:
@coderabbitai review

kane50613:
@coderabbitai review

PR #1133, opened 2026-08-06
PDF: write colours to four decimals

kane50613:
Stacked on #1132.

## Why

A colour component reaches the content stream as an eight-bit value divided by 255, and shortest-round-trip printing spells that out in full:

```
0.8352941 0.047058824 0.1764706 rg
```

Every painted element carries three of them.

## What

Four decimals. That resolves about eighty times finer than the 1/255 the value came from, so a reader converting back to eight bits lands on the same byte.

```
0.8353 0.0471 0.1765 rg
```

## Numbers

Against master, so these include #1130, #1131 and #1132:

| | before | after |
| --- | ---: | ---: |
| `invoice.pdf` | 34,390 | 23,306 (−32.2%) |
| 41 golden fixtures | 415,367 | 347,026 (−16.5%) |

Ghostscript renders every checked fixture byte-identically to master.

## Also tried

Rounding path coordinates and transforms to a thousandth of a point looked worth 5% of the uncompressed content stream, and was worth 16 bytes on the invoice once deflated: those long numbers are the same page transform repeated four hundred times, which LZ77 already collapses. Colours differ per element, which is why they pay. Not included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF color output by rounding color component values to four decimal places.
  * Updated color serialization across grayscale, RGB, CMYK, and spot-color formats for more consistent results.

* **Tests**
  * Updated PDF rendering checks to reflect the improved color precision for outlines, text shadows, and text strokes.

* **Documentation**
  * Added documentation describing the updated PDF color precision behavior.


kane50613:
@coderabbitai review

kane50613:
@coderabbitai review

PR #1134, opened 2026-08-06
PDF: convert each font once per renderer

kane50613:
Stacked on #1133.

## Why

Handing a font to the PDF writer does two things, both on every render:

- `Font::new` takes a `Data`, and `ShapedRun::font_data()` hands out a `&[u8]`, so `to_vec()` copies the whole blob. The blob behind it is `fontique::Blob`, which is reference-counted, but the borrow erases that.
- `Prehashed::new` then runs SipHash over the same bytes to give `Font` a cheap hash and equality.

On a 5 MB face those measure at 0.43 ms and 1.62 ms respectively, so the hash is the larger half by four times. A CJK invoice spent more time on the pair than on the rest of the render.

The raster and SVG backends do not pay this: they read glyphs through `FontRef::from_index` on a borrowed slice, with nothing owned and nothing hashed. Only krilla's `Font` owns its data.

## What

`PdfOptions` takes an optional `FontCache`, keyed the way the per-render map already was: the blob's stable id and its collection index. The wasm `PdfRenderer` keeps one beside its font context, so it lives and dies with the object JS already disposes.

Handing the blob's `Arc` through instead of copying would recover the smaller half and needs a `fontique` type in takumi-core's public API. The cache removes both halves without it.

## Numbers

| | before | after |
| --- | ---: | ---: |
| CJK invoice, 300 renders | 2.92 ms | 1.77 ms (−39.5%) |
| Latin invoice, 300 renders | 14.70 ms | 14.47 ms |
| wasm warm median, Latin | 26 ms | 26 ms |

The Latin gap is the whole cost of converting a 200 KB subset once. Both figures are native except the last.

Output is unchanged — no golden moved, which is the check that matters for a cache.

### DIFF
--- a/.tegami/pdf-font-cache.md
+++ b/.tegami/pdf-font-cache.md
@@ -1,0 +1,9 @@
+---
+packages:
+  takumi-pdf:
+    type: minor
+---
+
+### Convert each font once per renderer
+
+Converting a font for PDF output copies its whole blob and hashes it, and it was happening on every render. A `PdfRenderer` now keeps the converted fonts, so the second document onwards skips the work. A CJK invoice renders in 1.8 ms instead of 2.9 ms.
--- a/takumi-pdf-wasm/src/lib.rs
+++ b/takumi-pdf-wasm/src/lib.rs
@@ -18,7 +18,9 @@ use takumi_core::{
   resources::{font::FontResource, image::ResourceCache},
   style::{FontFamily, Lang},
 };
-use takumi_pdf::{Attachment, MeasureOptions, PdfMetadata, PdfOptions, PdfStandard, Tagging};
+use takumi_pdf::{
+  Attachment, FontCache, MeasureOptions, PdfMetadata, PdfOptions, PdfStandard, Tagging,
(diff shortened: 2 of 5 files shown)


kane50613:
@coderabbitai review

kane50613:
@coderabbitai review

kane50613:
Closing. Three things did not hold up:

The 39.5% figure came from a thread-local spike, not from this code. `PdfOptions` takes the cache, and the profiling example never passes one, so the shipped path was never the one measured.

Memory was never answered. A cached `Font` owns a copy of the whole blob, and the font registry already holds those bytes, so a renderer that draws with a large face would hold them twice for its lifetime. Sharing the blob's `Arc` instead would make the cache close to free, and it would want an accessor on `ShapedRun` that hands out the allocation without naming a parley type.

The builder does not take an optional cache cleanly either — `font_cache` is a typed-builder field, so passing one conditionally does not typecheck.

Worth redoing from the sharing angle: it removes the copy, the hash and the memory question together.

Issue #1176, opened 2026-08-09
takumi-pdf emits absolute-positioned zero-advance glyphs that lose word boundaries in Poppler


creosB:
## Summary

`takumi-pdf@0.4.2` produces selectable, tagged text that pdfjs reconstructs correctly, but Poppler `pdftotext` loses or invents word boundaries in ordinary prose. This matters for ATS/indexing pipelines because Poppler is a common independent extractor.

This does not look like a simple gap-threshold issue. Surviving and failing spaces in the same Inter run have the same physical gap ratio, while the content stream encodes every word and even each space as independently positioned text.

## Minimal reproduction

```tsx
import { writeFile } from "node:fs/promises";
import { googleFonts } from "@takumi-rs/helpers";
import { render } from "takumi-pdf";

const pdf = await render(
  <div style={{ fontFamily: "Inter", fontSize: 12.8, lineHeight: 1.47 }}>
    Built a hub-and-spoke architecture. Average App Rating. Full case studies: example.com/projects
  </div>,
  { size: "a4", fonts: await googleFonts([{ name: "Inter", weight: "400..700" }]) },
);
await writeFile("repro.pdf", pdf);
```

Then compare:

```bash
pdftotext -raw -enc UTF-8 repro.pdf -
pdftotext -layout -enc UTF-8 repro.pdf -
```

In a real two-page document generated from the same component/style path:

- pdfjs: intact reading order and word boundaries
- Poppler `-layout`: examples included `AppRating` and `hub-and-spokearchitecture`
- Poppler `-raw`: preserves `Built a hub-and-spoke architecture`, but badly fragments/merges other text, including `Average App Rating` and the space before a URL

## Gap controls

Measured from the rendered glyph boxes in the same PDF:

| Boundary | Gap | Font size | Gap/em | Poppler behavior |
| --- | ---: | ---: | ---: | --- |
| `Average App` | 2.7211pt | 9.675pt | 0.281 | merged/fragmented in `-raw` |
| `App Rating` | 2.7211pt | 9.675pt | 0.281 | merged/fragmented in `-raw` and `-layout` |
| `Built a` | 2.7000pt | 9.600pt | 0.281 | survives in `-raw` |
| `spoke architecture` | 2.7000pt | 9.600pt | 0.281 | survives in `-raw`, merges in `-layout` |
| `Local-first extraction` | 2.7000pt | 9.600pt | 0.281 | ordinary prose control |
[...]

PR #1177, opened 2026-08-09
PDF: read the fixtures back with Poppler

kane50613:
Investigates #1176 and adds the test that can prove it either way. It does not change rendering.

## Why

Nothing in the suite reads a generated PDF back as text. Conformance validation passes on a document whose words have merged, and so does every structure assertion, because the tag tree is right — it is the content stream a reader's extractor sees that is wrong.

Poppler is what an indexing or ATS pipeline most often runs, and it rebuilds words from glyph positions rather than from tags, so it catches a class of mistake nothing else here can.

## What

`poppler_reads_back_the_words` renders prose and asserts the phrases survive `pdftotext -raw` and `-layout`. It skips when `pdftotext` is missing, so a local checkout without Poppler still runs the suite; CI installs `poppler-utils` next to veraPDF, where it is not optional.

## What it found

**The proposed cause is not the cause.** `PdfGlyph::x_advance` returning `0.0` looks wrong, but krilla accumulates `x_advance - font_advance` and the next glyph's absolute `x_offset` cancels it exactly, so the emitted `TJ` is identical either way. Rewriting the glyphs to carry real advances moved 40 goldens by a few bytes and changed no extracted text, so it is not in this PR.

**Poppler discards the space character.** It drops `U+0020` and rebuilds spaces from the gap between glyph boxes, so a correct `ToUnicode` cannot save a space that measures too narrow.

**The threshold sits at about 0.2 em.** One document, one size, one family:

```
BoldAnnualreport
Plain Annual report
```

Archivo's space is 196 units at weight 700 and 209 at weight 400, measured from the font itself with fontTools. The PDF declares 196 for the bold subset, which is exactly right. Poppler keeps 0.209 and drops 0.196.

So this repository's own reproduction is a font metric landing under Poppler's threshold, not a defect in what takumi writes.

**That does not explain the report.** The gaps listed in #1176 that fail are 0.281 em, well above the threshold, and the content stream there resets `Tm` once per word, which none of these fixtures do. Reproducing it needs Inter and the run splitting that produced those separate text objects. The test added here is what that work should be measured against.

## Summary by CodeRabbit

* **Tests**
  * Added PDF text extraction validation to verify that generated documents preserve word boundaries and layout.
  * Added coverage for both raw and layout-based text output.

* **Chores**
  * Updated automated validation environments to support PDF text verification.
  * Tests gracefully skip optional checks when the required PDF utility is unavailable.


kane50613:
The Poppler oracle produced one red light in its life and it was a false positive: Archivo's bold space is 196/1000 em, under Poppler's ~0.2 em word-gap threshold, and the PDF declared it correctly. It did not catch the real defect behind #1176 either.

#1178 asserts the invariant directly instead: a shaped run emits one text-showing operation. No external binary, no dependence on another project's heuristics, and mutation-verified at 6 fragments versus 1.

PR #1196, opened 2026-08-09
PDF: give a link target something to point at under PDF/UA-2

kane50613:
## Why

A link to `#some-id` names a structure element rather than a place on a page, which is what ISO 14289-2 §8.8 asks of every link inside a document. Markup with nothing to say for itself, a plain `div` carrying only an id, produced no element, so the link named one that was never written. krilla falls back to a page destination without complaint, and neither its validator nor the `6.3.x` font sweep looks at this, so the render reported success and the file failed veraPDF at clause 8.8.

## What

A node a destination points at is tagged as a paragraph when it has no role of its own, and is kept even when it holds nothing: an element a link names is the one case where an empty structure element earns its place.

The test covers a target with text and an empty one, and every UA fixture still passes veraPDF, UA-2 included.

### DIFF
--- a/.tegami/pdf-ua2-link-destination.md
+++ b/.tegami/pdf-ua2-link-destination.md
@@ -1,0 +1,9 @@
+---
+packages:
+  takumi-pdf:
+    type: patch
+---
+
+### Give a link target something to point at under PDF/UA-2
+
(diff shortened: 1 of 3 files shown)


kane50613:
@coderabbitai review

PR #1210, opened 2026-08-10
Helpers, Core: fix <head> style collection, entity decoding and text-decoration: none

### LINKED ISSUE #1207: fromHtml drops <style> when it sits inside <head>
### Pre-submission Checklist

### Description

`fromHtml` collects `<style>` blocks into `stylesheets` when the input is a fragment, but returns an empty list when the same markup is a full document with a `<head>`. Every rule is dropped.

Since #1142, `render()` in `takumi-pdf` routes strings through `fromHtml` and joins the parsed `<style>` tags into that render's stylesheets, so the same input produces a completely unstyled PDF.

The reason this one cost us time is that it is **silent**. There is no warning and no error — the render succeeds and returns a valid, plausible-looking document that simply has no CSS applied. We were porting invoice templates and spent a while assuming our selectors were wrong.

### Steps to Reproduce

1. Call `fromHtml` with a `<style>` block and a `<div>`, as a fragment → `stylesheets.length === 1`
2. Wrap the exact same markup in `<html><head>…</head><body>…</body></html>` → `stylesheets.length === 0`

### Expected Behavior

A `<style>` element is collected wherever it legally appears, including inside `<head>`. Both calls should return one stylesheet.

### Actual Behavior

`stylesheets` is empty for the document form, and the rules are lost without any diagnostic.

### Reproduction Code

```js

const STYLE = `<style>.box{background:#2c82c9}</style>`;
const BODY = `<div class="box">x</div>`;

console.log(fromHtml(`${STYLE}${BODY}`).stylesheets.length);
// 1

console.log(
  fromHtml(`<!DOCTYPE html><html><head>${STYLE}</head><body>${BODY}</body></html>`)
    .stylesheets.length
);
// 0  <- expected 1
```

Same effect through `takumi-pdf`, where it is easier to see: the fragment renders a blue box, the document renders an unstyled one.

```js
import { render } from "takumi-pdf";

await render(`${STYLE}${BODY}`, { size: "a4" });                                   // styled
await render(`<html><head>${STYLE}</head><body>${BODY}</body></html>`, { size: "a4" }); // unstyled
```

### Affected Package

`@takumi-rs/helpers` (and `takumi-pdf` through it — not in the dropdown list yet)

### Version

`@takumi-rs/helpers` 2.7.0, `takumi-pdf` 0.6.2, Node v25.8.0, macOS arm64

### LINKED ISSUE #1208: fromHtml leaves HTML character references undecoded in text nodes
### Pre-submission Checklist

### Description

`fromHtml` copies HTML character references into text nodes verbatim instead of resolving them, so `&nbsp;`, `&deg;`, `&#176;` and `&quot;` are rendered as literal text.

This affects both named and numeric references, and it also hits values that were never authored as entities: a templating engine escaping an interpolated `"` into `&quot;` produces visible `&quot;` in the output.

#373 requested this and was closed, but the behaviour is still present in 2.7.0.

### Steps to Reproduce

1. Parse a `<div>` containing any character reference
2. Inspect the resulting node tree, or render it

### Expected Behavior

Character references are resolved during parsing, per the HTML spec, so the text node holds ` `, `°`, `"`.

### Actual Behavior

The text node holds the raw source: `a&nbsp;b &deg; &#176;`.

### Reproduction Code

```js

const { node } = fromHtml("<div>a&nbsp;b &deg; &#176; &quot;q&quot;</div>");
console.log(JSON.stringify(node));
// {"type":"text","text":"a&nbsp;b &deg; &#176; &quot;q&quot;","tagName":"div",...}
// expected: "a b ° ° \"q\""
```

### Affected Package

`@takumi-rs/helpers` (and `takumi-pdf`, which parses HTML strings through it)

### Version

`@takumi-rs/helpers` 2.7.0, `takumi-pdf` 0.6.2, Node v25.8.0, macOS arm64

### Additional Context

Working around it downstream is awkward: decoding the raw markup before `fromHtml` risks corrupting attribute values, so it has to be done by walking the parsed tree and rewriting every text node.

### LINKED ISSUE #1209: `text-decoration: none` is rejected and fails the whole render
### Pre-submission Checklist

### Description

`text-decoration: none` is rejected by the CSS parser. It is valid CSS — `none` is the initial value of `text-decoration-line` — but it fails the whole render rather than being ignored.

Because it aborts the render instead of dropping the declaration, a single `<a style="text-decoration:none">` anywhere in a document takes the entire PDF down. `text-decoration: none` on links is close to universal in existing HTML/CSS, so this is likely to hit anyone rendering pre-existing markup.

### Steps to Reproduce

1. Render any node carrying `text-decoration: none`, from an inline `style` attribute or a stylesheet
2. The call rejects

### Expected Behavior

`text-decoration: none` parses and resolves to no decoration.

### Actual Behavior

The render rejects with:

```
invalid value for textDecoration, line 1, column 1 near "none": string "none";
Unexpected token: none, expected a value of 'underline', 'line-through', 'overline',
'solid', 'from-font', <color>, <length> or <percentage>;
also accepts 'initial', 'unset' or 'inherit'.
```

### Reproduction Code

```js
import { render } from "takumi-pdf";

await render('<div style="text-decoration:none">x</div>', { size: "a4" });
// rejects
```

### Affected Package

`takumi` (Rust core, CSS parsing) — reproduced through `takumi-pdf`, not in the dropdown list yet

### Version

`takumi-pdf` 0.6.2, `@takumi-rs/helpers` 2.7.0, Node v25.8.0, macOS arm64

### Additional Context

Two separate things would each help on their own: accepting `none`, and having an unparseable declaration drop that declaration rather than fail the render.

kane50613:
## Why

A full HTML document lost every `<style>` rule because `fromHtml` skipped the `<head>` subtree (#1207). Character references like `&nbsp;` stayed as raw source in text nodes (#1208). `text-decoration: none` failed the whole render even though it is the initial value of `text-decoration-line` (#1209).

## What

`<head>` children are walked for stylesheets in both `fromHtml` and `fromJsx`, without emitting layout nodes for `<title>` and friends. Text nodes go through the same entity decoder as attributes, with the named table extended from 6 entries to the HTML4 set. The Rust parser accepts `none` as an empty `TextDecorationLines`.

Fixes #1207, fixes #1208, fixes #1209.

## Summary by CodeRabbit

* **New Features**
  * HTML and JSX inputs now collect styles from `<head>` elements without rendering head content.
  * Text and attribute values now decode standard, numeric, and hexadecimal HTML character references.
  * Unknown, invalid, or unsupported character references remain unchanged.

* **Bug Fixes**
  * CSS `text-decoration` and `text-decoration-line` now accept `none`, preventing rendering failures.
  * Invalid Unicode character references, including surrogate values, are safely preserved.


kane50613:
Skipping the `none` mutual-exclusivity suggestion: takumi's CSS value parsing is deliberately loose (first value wins, extra tokens tolerated), so `none underline` parsing as underline matches the rest of the parser. The other two comments are addressed in 9d36c7234.

PR #1215, opened 2026-08-12
Docs: render the playground browser preview in an iframe

kane50613:
## Why

The pane mounted the tree in a shadow root, where `rem` still resolves against the docs page's `<html>`. The engine resolves it against the tree's own outermost node, so any template that sets a font size on that node drew every `rem` length 1.5x off in one pane and not the other. Inherited properties crossed the shadow boundary too, so the docs dark theme painted the preview's text white.

## What

The pane renders into an iframe and hoists the outermost node's attributes onto that document's `<html>`, with `<body>` set to `display: contents`. `rem` now resolves against the same box the engine calls the root, and the frame's own document keeps the docs theme out.

An iframe never sizes itself to its content, so the flowing PDF layout measures `scrollHeight` and grows the frame. Both the measurement and the paint happen inside one `useLayoutEffect`, so the intermediate layout is never painted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser preview rendering for more accurate sizing and `rem` behavior.
  * Fixed style and font application within previews.
  * Improved support for both flowing and fixed-size preview layouts.
  * Set preview defaults to a light color scheme with consistent text and font styling.

### DIFF
--- a/.tegami/emoji-presentation.md
+++ b/.tegami/emoji-presentation.md
@@ -1,0 +1,9 @@
+---
+packages:
+  "@takumi-rs/helpers":
+    type: patch
+---
+
+### Follow Unicode emoji presentation in `extractEmojis`
+
+`extractEmojis` replaced every pictograph with a CDN image, so `‼` and `▶` came back as color emoji. Pictographs that default to text presentation now stay text, `U+FE0F` forces the emoji image, and `U+FE0E` forces the text glyph.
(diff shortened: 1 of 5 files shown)


kane50613:
Superseded by #1217: the engine now resolves `rem` against the viewport, so the shadow-root pane already agrees with the render. The theme leak that motivated the iframe is covered by the `color` pin in #1214.

PR #1277, opened 2026-08-14
Raster, SVG: clip an inset box-shadow to the padding box

flasyi:
An inset box-shadow was computed against the border box instead of the padding box. On a box with a border the shadow landed under the border, where the opaque border covered it, so no shadow appeared inside the box unless it was large enough to reach past the border. The shadow is now computed against the padding box and appears just inside the border, matching the PDF backend.

# Before / After
takumi-1786702168

# Reproduction Code

```
<!doctype html><html><head><title>inset box-shadow with a border</title></head><body style="margin:0;background-color:rgb(230,230,230)"><div style="box-sizing:border-box;width:240px;height:160px;background-color:rgb(255,255,255);border:18px solid rgb(120,140,200);border-radius:16px;box-shadow:inset 0px 0px 0px 30px rgba(0,0,0,0.6);" ></div></body></html>
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected inset box-shadow rendering so shadows appear inside the border, within the padding area.
  - Improved inset shadows for bordered and rounded elements, including more accurate clipping and positioning.
  - Prevented rendering issues for elements with non-positive dimensions.

- **Documentation**
  - Added release notes for the corrected inset shadow behavior.

### DIFF
--- a/.tegami/inset-shadow-padding-box.md
+++ b/.tegami/inset-shadow-padding-box.md
@@ -1,0 +1,13 @@
+---
+packages:
+  takumi-raster:
+    type: patch
+  takumi-svg:
+    type: patch
+  takumi:
+    type: patch
+---
+
+### Place an inset `box-shadow` in the padding box
+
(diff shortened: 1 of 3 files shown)


kane50613:
Thanks for the contribution!

PR #1326, opened 2026-08-17
fix(helpers): keep a `;` inside an inline style value

NgoQuocViet2001:
`parseInlineStyle` splits the `style` attribute on every `;`, but the character is legal inside a declaration value. The value is truncated at the first one and the remainder becomes a colon-less fragment the loop then discards.

The case that matters: a base64 data URL, which is how markup embeds an image without a fetch.

```
<div style="background-image:url(data:image/png;base64,iVBORw0KGgo=);color:red">
```

| input | before | after |
|---|---|---|
| `background-image:url(data:image/png;base64,iVBORw0KGgo=);color:red` | `{"backgroundImage":"url(data:image/png","color":"red"}` | `{"backgroundImage":"url(data:image/png;base64,iVBORw0KGgo=)","color":"red"}` |
| `font-family:'Foo; Bar', sans-serif;color:red` | `{"fontFamily":"'Foo","color":"red"}` | `{"fontFamily":"'Foo; Bar', sans-serif","color":"red"}` |
| `color:red;font-size:12px` | `{"color":"red","fontSize":"12px"}` | unchanged |
| `background-image:url(https://a.example/b.png)` | `{"backgroundImage":"url(https://a.example/b.png)"}` | unchanged |

So the image resolves to `url(data:image/png` and silently renders as nothing — the `base64,…` half is dropped without a warning. Colons were already handled through `valueParts.join(":")`, which is why the `https://` row survives and why this only shows up with `;`.

## The change

Declarations are split on top-level `;` only, tracking `url()` depth and quoted strings — everything else in `parseInlineStyle` is untouched:

```ts
for (const declaration of splitStyleDeclarations(styleText)) {
```

Backslash escapes inside a quoted string are skipped so `content:"a\";b"` stays one declaration, and an unbalanced `(` degrades to a single trailing declaration rather than losing the text.

This is the only such split in the repo — I grepped the TS and Rust sides. It reaches `fromHtml` / `fromStaticMarkup`, which is also the React DOM Server fallback for `fromJsx()`; the ordinary JSX path passes `style` as an object and never went through here.

## Tests

Four cases in `tests/html-markup.test.tsx`: the data URL, the quoted semicolon, ordinary declarations still separating (with a trailing `;`), and the unbalanced-`url(` fallback.

Negative control — reverting the loop to `styleText.split(";")` fails exactly the two new cases and leaves the other twelve green:

```
(fail) fromHtml > keeps a data URL intact in an inline style
(fail) fromHtml > keeps a quoted semicolon inside a declaration value
 12 pass
 2 fail
```

`bun test` on `takumi-helpers`, all files, on Windows:

```
emoji            11 pass  0 fail      element-types             1 pass  0 fail
html-markup      14 pass  0 fail      hooks-dispatcher          7 pass  0 fail
fonts            23 pass  0 fail      jsx-processing           42 pass  0 fail
renderer-facade   9 pass  0 fail      preact                    2 pass  0 fail
prepare-images    5 pass  0 fail      style-presets-override   27 pass  0 fail
                                      svg-serialize             6 pass  0 fail
```

`oxfmt --check` and `oxlint` clean at the pinned versions (0.63.0 / 1.78.0). Changelog added at `.tegami/inline-style-semicolon-split.md` as a patch to `@takumi-rs/helpers`.

One note: I could not run `tests/fetch-limits.test.ts` — it hangs on this machine with no output, which looks like it wants network. Untouched by this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inline style parsing to preserve semicolons inside `url()` values, quoted strings, and parenthesized values.
  * Correctly separates ordinary declarations and handles unclosed `url()` values without corrupting styles.
* **Tests**
  * Added coverage for data URLs, escaped and quoted semicolons, standard declaration separators, and incomplete URL values.
* **Documentation**
  * Added release documentation describing the inline style parsing improvements.

### DIFF
--- a/.tegami/inline-style-semicolon-split.md
+++ b/.tegami/inline-style-semicolon-split.md
@@ -1,0 +1,9 @@
+---
+packages:
+  "@takumi-rs/helpers":
+    type: patch
+---
+
+### Keep a `;` inside an inline style value
+
+`fromHtml` / `fromStaticMarkup` split the `style` attribute on every `;`, so a value carrying one of its own was truncated at the first and the remainder dropped. `style="background-image:url(data:image/png;base64,...)"` resolved to `url(data:image/png` — a data URL is the usual way markup embeds an image without a fetch. A quoted `font-family: "Foo; Bar"` was cut the same way. Declarations are now split on top-level `;` only, outside `url()` and quoted strings.
--- a/takumi-helpers/src/html/markup.ts
+++ b/takumi-helpers/src/html/markup.ts
@@ -235,22 +235,35 @@ function decodeAttributeMap(attributes: Record<string, string>): Record<string, 
  * Only a top-level `;` separates declarations. The character is legal inside a
  * value: `url(data:image/png;base64,...)` is how markup embeds an image without
  * a fetch, and a quoted family name such as `font-family: "Foo; Bar"` may carry
- * one too. Splitting on every `;` truncates the value at the first one and
- * leaves the remainder as a colon-less fragment that is then discarded.
+ * one too, as may an escaped `\;` outside either. Splitting on every `;`
+ * truncates the value at the first one and leaves the remainder as a
+ * colon-less fragment that is then discarded.
+ *
+ * A backslash escapes the next character in every state, which is how CSS
+ * treats it inside and outside a string alike.
  */
 function splitStyleDeclarations(styleText: string): string[] {
   const declarations: string[] = [];
   let start = 0;
   let depth = 0;
   let quote: string | undefined;
+  let escaped = false;
 
   for (let index = 0; index < styleText.length; index += 1) {
     const character = styleText[index];
 
+    if (escaped) {
+      escaped = false;
+      continue;
+    }
+
+    if (character === "\\") {
+      escaped = true;
+      continue;
+    }
+
     if (quote !== undefined) {
-      if (character === "\\") {
-        index += 1;
-      } else if (character === quote) {
+      if (character === quote) {
         quote = undefined;
       }
       continue;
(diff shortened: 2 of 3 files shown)


NgoQuocViet2001:
The escaped-semicolon finding is right — fixed in f5bab69.

I checked it before agreeing, because the first two ways I wrote the repro didn't actually test it: `\;` inside a JS template literal or a single-quoted string is an unrecognised escape, so it collapses to a bare `;` and the markup never carries a backslash at all. Building it with `String.fromCharCode(92)` gives a real one, and then the bug reproduces:

```
raw markup : "<div style=\"--value:foo\\;bar;color:red\">x</div>"
before     : {"--value":"foo\\","color":"red"}
after      : {"--value":"foo\\;bar","color":"red"}
```

`--value` was truncated to `foo\` and `bar` dropped, exactly as described.

The escape is now tracked before the quote and parenthesis states rather than inside the quoted branch, which matches CSS — a backslash escapes the next character outside a string as well as inside one. That also lets the quoted branch drop its `index += 1` special case, so there is one escape rule instead of two.

Value semantics are unchanged: the parser stores raw declaration values and does not unescape, so `foo\;bar` is kept with its backslash, consistent with how a quoted `'Foo; Bar'` keeps its quotes.

Regression test added, and it fails on the previous commit:

```
(fail) fromHtml > keeps an escaped semicolon outside quotes
 14 pass
 1 fail
```

Full `takumi-helpers` run stays green — 148 pass, 0 fail (html-markup now 15).

kane50613:
Superseded by #1371, which lands the same fix without the intermediate declaration array.

PR #1350, opened 2026-08-22
Core: play animated WebP image sources

### LINKED ISSUE #1337: Decoding an animated WebP `<img>` source throws "WebP encode failed"
### Pre-submission Checklist

### Description

When an animated WebP (VP8X container with `ANIM`/`ANMF` chunks) is used as an `` source, rendering fails outright with a decode error, instead of either rendering correctly or at least degrading gracefully to a static frame. This is more severe than the sibling APNG issue (silent first-frame-only rendering, #1336) — here the entire render request throws and produces nothing.

### Steps to Reproduce

1. Create a valid animated WebP with 3 distinct-colored frames (`VP8X` + `ANIM` + `ANMF`, 300ms/frame). We cross-validated this file independently with both `sharp`/`libvips` (reports a proper 3-page animated image with correct delays) before testing, so this is a genuine animated WebP, not a malformed one.
2. Use it as `` in a minimal HTML template, convert with `fromHtml()`.
3. Call either `renderer.render(node, options)` or `renderer.renderAnimation(...)`.

### Expected Behavior

Ideally the animation is consumed correctly. Failing that, at minimum it should degrade to rendering the first/default frame as a static image (i.e. the same — arguably still-buggy — behavior currently observed for animated PNG input, see the companion issue) rather than throwing and aborting the whole render.

### Actual Behavior

Both `render()` and `renderAnimation()` throw the same error:

```
Image resolution error: An error occurred while decoding the image data: Format error decoding WebP: WebP encode failed
```

To narrow this down we ran a few isolation tests:
- A single-frame plain `VP8`/`VP8L` WebP decodes fine.
- A single-frame WebP that still uses a `VP8X` container (e.g. for an alpha channel: `ALPH` + `VP8`) also decodes fine — so the `VP8X` container by itself isn't the problem.
- The failure only appears once the `VP8X` container also carries `ANIM`/`ANMF` (i.e. specifically when the file is animated).

We tested with animated WebP files produced by two independent encoders — `sharp`/libvips (both lossless VP8L and lossy VP8 variants) and `ffmpeg`'s `libwebp_anim` — all four combinations trigger the identical error, which rules out a single malformed/non-standard file being the cause.

### Reproduction Code

```ts

const renderer = new Renderer();
const html = `<div style="width:100%;height:100%;display:flex">
  <img src="${animatedWebpDataUri /* verified 3-frame animated WebP */}" style="width:100%;height:100%" />
</div>`;

const { node, stylesheets } = fromHtml(html);
// Throws immediately — both render() and renderAnimation() reproduce this:
await renderer.render(node, { width: 240, height: 190, format: "png", stylesheets });
// => Image resolution error: ... Format error decoding WebP: WebP encode failed
```

### Affected Package

takumi (Rust core), @takumi-rs/core, @takumi-rs/helpers

### Version

2.10.0

### Operating System

Windows

### Environment Details

- Bun: 1.3.14

### Playground Link

N/A — repro depends on local binary WebP assets.

### Additional Context

- The error message ("WebP encode failed") is also a little confusing for what's actually a *decode*-path failure — might be worth a look independent of the underlying animated-WebP support question.
- Because this errors instead of degrading, any pipeline that might receive animated WebP as an image asset (e.g. user-uploaded or CDN-sourced decorative images) needs to pre-validate/transcode it before handing it to Takumi, otherwise a single bad asset takes down the whole render request rather than just losing its animation. Might be worth calling out explicitly in `/docs/load-images` alongside (or instead of) documenting the APNG limitation, since the failure mode here is sharper than "static formats are decoded as their first frame" would suggest.
- Full disclosure: same as the companion APNG issue — this investigation was carried out with AI assistance (Claude), test assets were independently cross-validated (via `sharp`/libvips metadata plus a manual RIFF chunk walk) across two different encoders before concluding this is a real bug. This GitHub account is on loan from the person actually running the investigation; happy to share the full repro project if useful.

kane50613:
Fixes #1337

## Why

On non-wasm targets WebP decodes through libwebp's `WebPDecode`, which does not handle a `VP8X` container carrying `ANIM`/`ANMF`. Every animated WebP failed the whole render, reporting `WebP encode failed` from a decode path. Nothing checked whether the file was animated.

## What

Animated WebP decodes through `image-webp`, which owns the animation canvas, and plays on the same timeline as GIF. Still WebP keeps libwebp, including its direct scaled decode. `decode_image`, which the SVG backend and the size cache go through, falls back to the first frame instead of failing, and `decode_webp_scaled` declines animated bytes so a later routing change cannot put them back on libwebp. The libwebp status now reports as invalid encoded data.

`GifSource` becomes `AnimatedSource` and `ImageSource::Gif` becomes `ImageSource::Animated` in a separate commit that changes no behavior. Frame decoding, timing scans and the frame budget are shared; per-format work stays behind a private enum rather than a trait.

Frames past 1024 are dropped: the pixel budget alone lets a tiny canvas carry an unbounded frame list.


kane50613:
@coderabbitai review

kane50613:
Pushed a round of fixes:

- `webp_frame_durations` and `apng_frame_durations` now stop on the same frame and pixel budgets the decoders use, so a playback time can no longer select a frame the decoder drops. Thanks — this one was real, and the APNG scan had it too.
- Fixture colors moved to a format-neutral `FRAME_COLORS` without the `gif` gate. `--no-default-features --features webp` did not compile before; it does now, alongside gif-only and no-feature builds.
- `ImageError::InvalidGif` is now `ImageError::InvalidAnimation`.

Not taking the fallback-to-`decode_image` suggestion for a malformed animated input. Both issues behind this stack were reported as *silent* wrong output, and quietly turning a broken animation into a still image is the same failure mode again — a decode error the caller can see is the better outcome. A file that only fails the animation path would almost certainly fail `decode_image` too.

kane50613:
Two of the three taken:

- The frame decode no longer swallows its error. A failure now logs which frame failed and why, and the both-decodes-failed fallback logs before drawing the transparent pixel — a blank image is traceable instead of silent.
- The SVG raster path takes the decoded buffer via `Arc::unwrap_or_clone` + `into_premultiplied_rgba` instead of copying the canvas.

On the O(N²) sampling: measured rather than argued. A 120x120 source rendered into an output animation of the same length, release build, whole render pipeline:

| source frames | output frames | time |
| --- | --- | --- |
| 3 | 3 | 0.87ms |
| 30 | 30 | 11.7ms |
| 120 | 120 | 109ms |

The quadratic term is real and visible in those numbers, but 120 frames costs a tenth of a second end to end, which is not worth a thread-local decode cursor and its invalidation rules yet. Leaving the `ponytail:` note in place — it already names the ceiling and the upgrade path, so whoever hits it has the fix written down.

kane50613:
Fixed the build break `takumi-pdf` hit: `log` was only pulled in by the `svg` feature, and the two warnings I added in this layer are not svg-specific. It is now a plain dependency of `takumi-core` rather than an optional one — it is a facade crate, so this costs nothing in a build that never logs.

Verified across `--no-default-features`, gif-only, webp-only, `--all-features`, and a `takumi-pdf-wasm` wasm32 build.

PR #1351, opened 2026-08-22
Core: play APNG image sources

### LINKED ISSUE #1336: Animated PNG `<img>` source only renders its first frame — animation is silently dropped
### Pre-submission Checklist

### Description

When an animated PNG (APNG, with `acTL`/`fcTL`/`fdAT` chunks) is used as an `` source and rendered via `renderAnimation()`, Takumi only reads the APNG's default/first frame and renders that single frame across the entire output animation — the source animation is silently discarded. No error or warning is raised, so this is easy to miss: the render "succeeds" but the visual result has lost all animation.

### Steps to Reproduce

1. Create a valid 3-frame APNG (`acTL` frame_count=3, `fcTL` delay=300ms each frame, three visually distinct frames — we used solid red/green/blue). We cross-validated the test file with an independent decoder (`upng-js`) before testing, confirming 3 real frames with correct per-frame RGBA pixel data and delay — this is a genuine multi-frame APNG, not a mislabeled static file. (Note: `sharp`/`libvips` itself reports `pages: undefined` for this same file — it does not support APNG decoding at all, so it can't be used to validate APNG test assets; `upng-js` or an equivalent APNG-aware decoder is required.)
2. Build a minimal HTML string whose only content is `` (the APNG) filling the render canvas, and convert it with `fromHtml()`.
3. Call `renderer.renderAnimation({ fps: 10, format: "gif", scenes: [{ node, durationMs: 900 }] })`.
4. Decode the resulting output animation's frames (we wrote a small dependency-free GIF frame/delay parser for this) and sample a pixel from each frame.

### Expected Behavior

The output animation should cycle red → green → blue over time (sampled according to the requested `fps`/`durationMs`), reflecting the source APNG's declared per-frame timeline.

### Actual Behavior

All 9 output frames (10fps × 900ms) are pixel-identical, showing only the APNG's first/default frame (red). Green and blue never appear anywhere in the output. No error, warning, or any other signal is raised — the call completes "successfully" with a wrong result.

We also ran the same test with a normal (non-APNG) animated GIF as a control, using the identical HTML/render code path — the GIF case correctly preserves all 3 colors across the output frames. So this appears to be specific to APNG input, not a general problem with our test harness.

### Reproduction Code

```ts
// Simplified from our repro (Bun + @takumi-rs/core + @takumi-rs/helpers 2.10.0)

const renderer = new Renderer();
const html = `<div style="width:100%;height:100%;display:flex">
  <img src="${apngDataUri /* verified 3-frame APNG, red/green/blue, 300ms each */}" style="width:100%;height:100%" />
</div>`;

const { node, stylesheets } = fromHtml(html);
const gif = await renderer.renderAnimation({
  fps: 10,
  format: "gif",
  scenes: [{ node, durationMs: 900 }],
  stylesheets,
});
// Decoding `gif` frame-by-frame: all 9 frames are RGB(255,0,0) — never green, never blue.
```

### Affected Package

takumi (Rust core), @takumi-rs/core, @takumi-rs/helpers

### Version

2.10.0

### Operating System

Windows

### Environment Details

- Bun: 1.3.14
- (No browser / Node.js involved — pure Bun runtime)

### Playground Link

N/A — repro depends on a local binary APNG asset that isn't practical to paste into the hosted Playground.

### Additional Context

- Source-level, for what it's worth (we're not deeply familiar with the codebase, so take this as a pointer rather than a diagnosis): `takumi-core/src/resources/image_decoder.rs`'s `decode_png` appears to always go through `PngDecoder` + `DynamicImage::from_decoder`, i.e. a single default-frame decode, with nothing reading `acTL`/`fcTL`/`fdAT` for a per-frame timeline — consistent with what we observed. `GifSource`/`frame_at_time` in `image.rs` appears to be the only input path that's actually time-axis aware.
- We looked through the docs (`/docs/reference`, `/docs/load-images`, `/docs/output-formats`, `/docs/integration`) and didn't find this called out anywhere. The `load-images` "Decode limits" table only lists a special multi-frame budget for "Animated GIF," which hints at single-frame handling for everything else, but nothing states it explicitly.
- Because this fails *silently*, it's a bit of a footgun for anyone using animated PNG assets for backgrounds/decorations (that was our use case) — the render doesn't error, it just quietly produces the wrong image. Even before any full APNG-decoding support, a warning/log line when a multi-frame-capable input (PNG with `acTL`, or animated WebP — see the companion issue #1337) is reduced to its first frame would probably save people a lot of confusion.
- Full disclosure: this investigation was carried out with AI assistance (Claude), including generating and independently cross-validating the test assets (via `upng-js`, plus a second APNG produced with a different encoder/ffmpeg to rule out a single-tool artifact) before concluding this is a real gap rather than a test-file mistake. This GitHub account is being used on loan from the person actually running the investigation — happy to share the full repro project if that's useful for triage.

kane50613:
Fixes #1336

## Why

`decode_png` only ever ran a single-frame decode, so an APNG's `acTL`/`fcTL`/`fdAT` chunks were never read. A render held the default image for the whole animation and raised nothing to say the rest had been dropped.

## What

APNG joins the animated sources added one layer down. Frames decode in stream order and composite through the `fcTL` dispose and blend operations. Per-frame delays come from a chunk scan that decodes no pixels, so the timing table and the decodable frames stop at the same point. A default image that no `fcTL` claims is decoded past and never enters the timeline, rather than being given an invented duration.

The render fixture drives the same source frames through WebP and APNG.


kane50613:
@coderabbitai review

kane50613:
Good catch on the blend coverage. The subframe is now half-transparent blue over opaque red, so the alpha-weighted branch of `blend_over` actually runs; the expected value is worked out by hand from source-over rather than recorded from the output.

kane50613:
Fixed — the cap counted reads, not timeline slots, so an unclaimed default image cost one animation frame. The loop now allows one read past `MAX_ANIMATION_FRAMES` and caps on frames that actually enter the timeline, which also keeps it consistent with `apng_frame_durations` counting `fcTL` chunks.

Verified differentially with the cap temporarily lowered to 2: the old read-counted loop returned only the first animation frame for a default-image-plus-two-frames APNG, the new one returns both. Left behind a regression test for the default image staying off the timeline (built with `set_sep_def_img`); skipped building a 1024-frame fixture, since the cap arithmetic is what changed and a full-size fixture would only slow the suite down.

kane50613:
Half of this one was real: the duplicated comment was mine, from restoring the loop after a differential check. Removed.

The double-advance concern does not hold, though. `next_frame_info()` calls `read_until_image_data()` and leaves `subframe.consumed_and_flushed` false; `next_frame()` only advances when that flag is true, so it decodes exactly the frame `next_frame_info()` just positioned on. Checked against a four-frame APNG with distinct per-frame values — the decoder emits all four in order, none skipped.

The call is also required rather than incidental: `output_buffer_size()` reflects the current subframe's dimensions, so without advancing to the next `fcTL` first, the buffer gets sized from the previous frame.

kane50613:
Skipping this one. The `clone` is not scratch space: `ImageBuffer::from_premultiplied_rgba` takes the `Vec` by value and it becomes the returned buffer's storage, so every emitted frame needs its own allocation regardless. A reusable field would be moved out at that same call and reallocated for the next frame, while permanently holding a canvas-sized buffer — the opposite direction from dropping the retained first frame one layer down.

The resample path could in principle avoid one allocation, but a draw emits a single frame (intermediate frames never reach `to_buffer`), so there is no per-frame allocation loop to tighten here.

PR #1353, opened 2026-08-22
Core: start backwards-filled step animations on the first step

kane50613:
## Why

The step easing algorithm needs to know whether a sample was taken before the animation's active phase: a sample sitting exactly on a jump belongs to the step below it. Nothing carried that flag, so an animation with `animation-fill-mode: backwards` or `both` and a positive delay held the wrong value throughout the delay. `steps(4, jump-start)` sat at 0.25 and `steps(4, jump-both)` at 0.2, where both should have been 0.

Serialization had a second, smaller gap: an end position is the default, so `steps(4, end)` and `steps(4, jump-end)` are supposed to print as `steps(4)`.

## What

`sample_animation_progress` returns an `AnimationSample` carrying the progress and a `before` flag, which `apply_timing_function` passes to the step sampler; on an integral step boundary a before-sample takes the step below, following the algorithm in css-easing-1. Other timing functions ignore the flag, as they should. `to_css` leaves out an end position.

Found by an adversarial review of #1346.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed step-based animations with positive delays so backwards-filled states display the correct preceding step.
  * Corrected sampling at exact step boundaries during pre-animation phases.
  * Improved serialization of `steps()` timing functions by omitting the default end position.

### DIFF
--- a/.tegami/steps-before-flag.md
+++ b/.tegami/steps-before-flag.md
@@ -1,0 +1,11 @@
+---
+packages:
+  "@takumi-rs/core":
+    type: patch
+---
+
+### Start backwards-filled step animations on the first step
+
+An animation with `animation-fill-mode: backwards` or `both` and a positive delay held the second step during the delay when its timing function was `steps(n, jump-start)` or `steps(n, jump-both)`. Samples taken before the active phase now land on the step below the jump, as the CSS easing algorithm requires.
+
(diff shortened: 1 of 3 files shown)


kane50613:
@coderabbitai review

kane50613:
@coderabbitai review

PR #1355, opened 2026-08-22
Core: decode a standalone animated frame without replaying the timeline

kane50613:
## Why

Frame disposal is stateful, so reaching frame N meant compositing 0..N. Sampling an animation at every frame was quadratic, and a long source rendered into an output animation ran into tens of seconds.

Most frames do not need that. A frame covering the whole canvas that replaces what is under it has no dependency on its predecessors, which is the sprite-sheet shape both animated image issues were about.

## What

Each format answers "does this frame stand alone?" from container metadata, without decoding pixels, then decodes it directly when it does:

- WebP walks the RIFF chunks to the Nth `ANMF` and lifts its bitstream into a still container. A lossy frame with a separate `ALPH` chunk gets a `VP8X` header rebuilt, since only the animation container carried one. The decoded frame's real dimensions are checked against the rectangle the `ANMF` claimed, so a header that overstates its bitstream goes back to replay instead of drawing at the wrong size.
- APNG advances by `next_frame_info()` and decodes the frame it lands on.
- GIF advances the same way without reading pixels into a buffer.

Dependent frames still replay. Blink decides this in `ImageDecoder::FindRequiredPreviousFrame` and reaches WebP frames through libwebp's `WebPDemuxGetFrame`; walking RIFF by hand keeps it working on wasm, where that demuxer is absent.

`ApngCanvas::composite` copies whole rows when RGBA pixels replace what they land on, and the standalone APNG path skips the compositing surface, since a full-canvas replacing frame already is the canvas.

Sampling the last frame of a 300-frame 200x200 animation:

| source | before | after |
| --- | --- | --- |
| standalone WebP | 96.8ms | 0.049ms |
| standalone GIF | 16.6ms | 0.128ms |
| standalone APNG | 158.5ms | 0.387ms |
| dependent APNG | 38.6ms | 38.6ms |

The last row is why the benchmark has a `dependent` group: nothing here helps a frame that genuinely blends onto its predecessor. Caching the nearest ancestor such a frame needs, as `ClearCacheExceptTwoFrames` does, is the next step and is not in this PR.

Three tests guard the risk that a frame looks standalone but is not: seek and replay must produce byte-identical pixels for the same frame, a partial-canvas `BlendOp::Over` frame must decline to seek, and an `ANMF` claiming the full canvas while carrying a smaller bitstream must decline too.

### DIFF
--- a/.tegami/animated-frame-seek.md
+++ b/.tegami/animated-frame-seek.md
@@ -1,0 +1,9 @@
+---
+packages:
+  "@takumi-rs/core":
+    type: patch
+---
+
+### Draw animated frames without replaying the ones before them
+
+A frame that covers the whole canvas and replaces what is under it does not depend on the frames before it, so it is now decoded on its own. Reaching the last frame of a 300-frame animation drops from 16.6ms to 0.13ms for GIF, 158ms to 0.39ms for APNG, and 97ms to 0.05ms for WebP. Frames that do blend onto their predecessors still replay, as they must.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1795,7 +1795,9 @@ version = "2.10.0"
 dependencies = [
  "criterion",
  "image",
+ "image-webp",
  "parley",
+ "png",
  "rayon",
  "resvg",
  "serde_json",
--- a/takumi-core/src/resources/image.rs
+++ b/takumi-core/src/resources/image.rs
@@ -24,7 +24,8 @@ use xxhash_rust::xxh3::{Xxh3, xxh3_64};
 use crate::resources::image_decoder::decoder_compiled_out;
 #[cfg(feature = "webp")]
 use crate::resources::image_decoder::{
-  animated_webp_dimensions, decode_webp_frames, is_animated_webp, webp_frame_durations,
+  animated_webp_dimensions, decode_webp_frame_alone, decode_webp_frames, is_animated_webp,
+  webp_frame_durations,
 };
 #[cfg(feature = "svg")]
 use crate::resvg::{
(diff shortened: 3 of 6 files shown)


kane50613:
Two fixes:

- `decode_gif_frame_alone` was missing its feature-off counterpart, so a build without the `gif` feature could not resolve the import. That is what broke `Build takumi-pdf`; the other gif entry points all have one and this one should have too.
- The three seek paths now reject an index at or past `MAX_ANIMATION_FRAMES` before scanning, as suggested. The only current caller derives the index from the timing table, which is already capped, so this changes no behavior today — it keeps the bound with the function rather than with its caller.

Checked against `--no-default-features`, gif-only, webp-only, `--all-features`, and `takumi-pdf-wasm` on wasm32.

kane50613:
Ran a simplify pass over this layer (four parallel reviews: reuse, simplification, efficiency, altitude) and applied what held up:

- The seek paths return `Option<ImageBuffer>` instead of `Option<Result<..>>`. Every real decode failure was already becoming `None`; the only reachable `Some(Err)` was a resample failure, which made `decode_frame` treat it as fatal and skip the replay that would have failed identically. `None` now uniformly means "no shortcut, go replay", and the duplicated warn is gone.
- `decode_webp_frame_alone` no longer calls `animated_webp_dimensions`. Constructing a `WebPDecoder` walks every `ANMF`, builds a HashMap and seeks per frame just to read two `u32`s — the `VP8X` chunk carries the canvas size, and the walk to the Nth `ANMF` was happening anyway. Standalone WebP at 300 frames: **80us to 49us**, and the O(frames) term is gone.
- `MAX_ANIMATION_FRAMES` is checked once in `AnimatedFormat::decode_frame_alone` rather than three times in the format decoders.
- `covers_canvas` is shared; the three formats keep their own "replaces rather than blends" test, since that genuinely differs per container.
- `webp_canvas_to_buffer`'s tail was `fit_to_target` written out; the APNG walk loop was a `for` loop with an extra idle iteration; the two APNG bench builders were one builder with a flag.
- Both chunk walkers use checked arithmetic, matching the two RIFF walkers already in takumi-raster.
- Comment pass per the repo's policy: dropped the ones restating code, kept the format-spec invariants (GIF alpha being 0 or 255, the unclaimed default image, the ANMF flag bit).

Left for later, with measurements: the APNG header walk is ~48% of its seek path and could be replaced by locating the `fcTL` via chunk scan and rebuilding a one-frame PNG (372us to ~200us, and O(1) in frame count); and the `opt-level = "z"` overrides on `png`/`takumi-core` predate this code decoding per draw — worth ~11% on the dependent path, but that is a size-versus-speed call for the repo to make, not something to slip into this PR.

PR #1358, opened 2026-08-22
Core: rebuild an APNG frame instead of walking to it

kane50613:
## Why

Reaching frame N of an APNG advanced with `next_frame_info()`. That is not a cheap skip: png reads and checksums every preceding frame's `fdAT` bytes. At 300 frames it was about half the seek path, and it grew with frame count.

## What

Each APNG frame carries its own complete zlib stream, so a frame can be rebuilt as a standalone PNG: the original `IHDR` with the frame's dimensions, the palette chunks, the frame's data as `IDAT`, and `IEND`. `png_chunks` already walks by pointer, so locating it is nearly free.

The old walk stays as the fallback.

| frames | before | after |
| --- | --- | --- |
| 30 | 214us | 87us |
| 120 | 246us | 104us |
| 300 | 250us | 98us |

Ten times the frames now costs 26% more instead of scaling with the stream.

Chunk checksums are computed a bit at a time rather than pulling in `crc32fast`. One rebuild runs it once, and measuring both ways showed the same speed and the same bundle size.

A test asserts a seeked subframe is byte-identical to replaying to it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved APNG decoding for independently seekable subframes.
  * Fixed rendering of positioned APNG frames on cleared canvases.
  * Preserved expected pixel output when seeking to animation frames.
  * Improved support for APNG frame scaling and disposal behavior.
  * Added fallback handling when standalone frame reconstruction is unavailable.
* **Tests**
  * Added regression coverage for APNG frame seeking and pixel consistency.

### DIFF
--- a/takumi-core/src/resources/image.rs
+++ b/takumi-core/src/resources/image.rs
@@ -1803,6 +1803,47 @@ mod tests {
     assert_eq!(gif.frame_at_time(250).pixel(2, 2), FRAME_COLORS[2]);
   }
 
+  /// Rebuilding a subframe as a still PNG has to swap `IHDR` to the frame's own
+  /// size, then place it back at its offset.
+  #[test]
+  fn a_seeked_apng_subframe_matches_replaying_to_it() {
+    use png::{BitDepth, ColorType, DisposeOp, Encoder};
+
+    let mut bytes = Vec::new();
+    let mut encoder = Encoder::new(&mut bytes, 4, 4);
+    encoder.set_color(ColorType::Rgba);
+    encoder.set_depth(BitDepth::Eight);
+    encoder.set_animated(3, 0).unwrap();
+
+    let mut writer = encoder.write_header().unwrap();
+    for (index, color) in FRAME_COLORS.iter().enumerate() {
+      writer.set_frame_delay(100, 1000).unwrap();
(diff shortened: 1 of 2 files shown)


kane50613:
Skipping this one. `data` accumulates `IDAT` and `fdAT` payloads out of `bytes`, and `png_chunks` advances monotonically through that slice, so the total cannot exceed the input the caller already handed us. Peak is bounded at roughly twice the input, which is the same shape as any other decode path here.

The unbounded growth that would matter is after inflate, and that is governed by the png decoder's own limits rather than by this copy.

kane50613:
Closing as micro-optimization.

250us to 98us on a 300-frame APNG, which nobody is rendering — the issues behind this work were sprite sheets of a handful of frames, where seeking is already microseconds after #1357.

It also has a poor risk profile for what it buys. Reviewing it here surfaced two real defects in the hand-rebuilt PNG: a short `fcTL` indexes past its payload and panics, and a frame rectangle reaching outside the canvas takes the same-size fast path and lands in the wrong place. Both are introduced by this PR and neither exists on master. Reconstructing a container by hand earns that kind of bug, and the speed does not pay for it.

The branch stays if a long APNG ever makes this worth revisiting.

PR #1366, opened 2026-08-23
Core: match selectors against the rules that could apply

kane50613:
## Why

Every node ran every selector of every media-active rule. The ancestor bloom filter helps only with combinators; a rule whose subject is `.some-class` still had its matcher run against every node in the document.

That is `O(nodes x rules)`, and nothing measured it: the benchmarks render small trees against small stylesheets. Against a utility framework, where a node matches a handful of several hundred rules, it dominates.

| nodes x rules | before |
| --- | --- |
| 200 x 200 | 4.48ms |
| 400 x 400 | 17.2ms |
| 800 x 800 | 67.3ms |

Four times the input costs fifteen times the work.

## What

Rules are grouped by what their rightmost compound requires — an id, a class, a tag, or nothing indexable — and a node collects the buckets matching its own id, classes and tag, plus the unindexable ones. Source order is preserved by merging bucket contents and sorting, so the cascade is unchanged.

A rule whose selector list contains anything unindexable goes wholly into the unindexed bucket, so a rule is never skipped when one of its selectors could have matched.

| nodes x rules | after | |
| --- | --- | --- |
| 200 x 200 | 0.52ms | 8.7x |
| 400 x 400 | 1.09ms | 15.8x |
| 800 x 800 | 2.49ms | 27x |

Growth now follows the rules that could match rather than the size of the stylesheet.

The `cascade` benchmark is new, and is the point of this PR as much as the change is: nothing in the suite scaled the stylesheet before, which is why this went unnoticed.

Every fixture is byte-identical and the cascade tests pass unchanged, which is the check that matters — this must not alter which declarations win.

## Summary by CodeRabbit

  * Improved style rule matching to evaluate relevant rules more efficiently.
  * Preserved styling behavior and rule order while reducing unnecessary selector checks.
  * Added performance coverage for cascading styles across different document and stylesheet sizes.

* **Documentation**
  * Added guidance describing selector matching behavior and reported performance results.

### DIFF
--- a/.tegami/selector-index.md
+++ b/.tegami/selector-index.md
@@ -1,0 +1,9 @@
+---
+packages:
+  "@takumi-rs/core":
+    type: patch
+---
+
+### Match selectors against the rules that could apply
+
+Stylesheet rules are grouped by what their rightmost selector requires, so a node only runs the matcher against rules naming one of its classes, its id or its tag. Rendering 800 nodes against 800 rules drops from 67ms to 2.5ms; the cost now grows with the rules that could match rather than with the whole stylesheet.
--- a/takumi-core/src/matching.rs
+++ b/takumi-core/src/matching.rs
@@ -11,7 +11,7 @@ use selectors::{
   attr::{CaseSensitivity, NamespaceConstraint},
   bloom::BloomFilter,
   matching::*,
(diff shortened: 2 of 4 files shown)


kane50613:
@coderabbitai review

PR #1367, opened 2026-08-23
Core: reuse layout styles that do not depend on their query container

kane50613:
## Why

Taffy calls `compute_child_layout` several times per node — min-content, max-content, and the flex passes. Every call rebuilt that node's `taffy::Style` from scratch, allocating grid template vectors and re-resolving every length.

The only reason to rebuild is `container_size`. The only things that read it are container query units, and most styles use none.

## What

Length resolution records whether it read the query container. A style that never read one is built once and reused; a style that did is rebuilt as before.

| nested flex | before | after |
| --- | --- | --- |
| depth 4 (~340 nodes) | 3.70ms | 2.18ms |
| depth 5 (~1.4k nodes) | 13.8ms | 7.90ms |
| depth 6 (~5.5k nodes) | 59.0ms | 36.8ms |

The `layout` benchmark is new. The suite had no deeply nested tree, which is why this went unmeasured.

An earlier version resolved the style twice against different containers and compared. That is not sound: `min(10px, 100cqw)` agrees across two large containers and disagrees with a small one. Recording the read is exact, and stays exact if a new property starts consulting the container.

Every fixture is byte-identical, including the container query ones.

## Summary by CodeRabbit

* **Performance**
  * Improved layout performance for deeply nested flex structures by reusing styles when container-query sizing does not affect them.
  * Reduced unnecessary style recalculation during repeated layout passes.

* **Benchmarks**
  * Added coverage for nested flex layouts across multiple tree depths and configurations.

* **Documentation**
  * Added a patch-release note describing the layout performance improvement.

### DIFF
--- a/.tegami/taffy-style-reuse.md
+++ b/.tegami/taffy-style-reuse.md
@@ -1,0 +1,9 @@
+---
+packages:
+  "@takumi-rs/core":
+    type: patch
+---
+
+### Reuse layout styles across sizing passes
+
+Layout runs several sizing passes over each node, and each pass rebuilt that node's layout style from scratch. Only container query units (`cqw`, `cqh`, `cqmin`, `cqmax`) can change between passes, so a style using none of them is now built once and reused. A deeply nested flex tree lays out about 36% faster.
--- a/takumi-core/src/layout/tree.rs
+++ b/takumi-core/src/layout/tree.rs
@@ -99,6 +99,9 @@ pub struct LayoutTree<'r> {
 
 struct LayoutNodeState {
   style: Style,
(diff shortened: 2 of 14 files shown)


kane50613:
Good catch — that was unsound, and would have bitten as soon as `min()` or `clamp()` landed.

Replaced the comparison with an exact signal: length resolution records whether it read the query container, and the style is reused only when nothing did. A value that resolves the same against every container still reports the read, since the flag tracks what was consulted rather than whether it mattered.

It also turned out faster than the version you reviewed, because it drops the second resolve: depth 6 is now 36.8ms against 59.0ms on master, where the compared version measured 37.8ms.

PR #1384, opened 2026-08-24
Core: customize the tw prop's design tokens

kane50613:
## Why

The theme resolution added below this layer had no way in: the docs told people the `tw` prop has no theme config, and a project's brand color or spacing step had to be spelled as an arbitrary value on every element.

## What

Renders take a `theme` option keyed by CSS custom property name, alongside `stylesheets` and `keyframes`. It rides on the `StyleSheet` the render already carries, so the napi, wasm and PDF bindings all share one lowering. `--color-brand-500` makes `bg-brand-500` work, `--spacing-gutter` makes `p-gutter` work, and `--color-red-500` replaces the built-in red.

The option is a flat record rather than a nested config so it stays the same shape as the CSS custom properties Tailwind itself compiles a theme down to. Object key order carries no meaning in JS, so namespace resets are applied before the tokens in the same batch.

Both layers together cost 3.6 KB gzipped on the shipped wasm pipeline (1,556,162 to 1,559,846 bytes, local build with the CI toolchain and wasm-opt flags).


kane50613:

PR #1386, opened 2026-08-24
Core: cover every Tailwind theme namespace the parser can reach

kane50613:
## Why

Only `--color-*` and `--spacing-*` reached a utility, so a project could theme its palette and spacing step but not its type scale, radii, shadows or animations.

## What

Fifteen namespaces are wired up, every one Tailwind documents that has a utility here to reach: colors, spacing, containers, `--text-*`, `--font-*`, `--font-weight-*`, `--tracking-*`, `--leading-*`, `--radius-*`, `--shadow-*`, `--drop-shadow-*`, `--text-shadow-*`, `--blur-*`, `--aspect-*` and `--animate-*`. Each is one associated const on the value type. The remaining upstream namespaces (`--breakpoint-*`, `--ease-*`, `--inset-shadow-*`, `--perspective-*`, `--zoom-*`, `--tab-size-*`) have no matching utility in the parser.

Two cases broke the "one namespace list per value type" rule and get a per-candidate override in `map.rs`: `TextShadow` serves both `--text-shadow-*` and `--drop-shadow-*`, and `Length` reads `--container-*` for `max-w-*` where it reads `--spacing-*` everywhere else. Coverage is asserted by equivalence: a theme token must resolve to the same declarations as the matching arbitrary value, so `text-hero` with `--text-hero: 4rem` has to match `text-[4rem]`.

## Summary by CodeRabbit

* **New Features**
  * Expanded theme token support across typography, sizing, containers, borders, shadows, blur, aspect ratios, and animations.
  * Added namespace-aware utility resolution, including container-specific sizing tokens.
  * Improved precedence when matching overlapping theme tokens.

* **Documentation**
  * Updated styling documentation with supported namespaces, utility mappings, and unsupported namespace details.

* **Tests**
  * Added coverage confirming themed utilities match equivalent arbitrary values across supported styling categories.

### DIFF
--- a/.tegami/tw-theme.md
+++ b/.tegami/tw-theme.md
@@ -12,4 +12,4 @@ packages:
 
 The built-in Tailwind parser only knew Tailwind's own scales, so a project's brand color or spacing step had to be spelled as an arbitrary value on every element. Renders now take a `theme` option keyed by CSS custom property name: `--color-brand-500` makes `bg-brand-500` work, `--spacing-gutter` makes `p-gutter` work, and naming a built-in token such as `--color-red-500` replaces it.
 
-A value of `initial` removes a token, and `--color-*: initial` clears the namespace along with the built-in palette underneath it. Only `--color-*` and `--spacing-*` are read so far.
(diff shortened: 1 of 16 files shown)


kane50613:

PR #1388, opened 2026-08-24
Core: let a stylesheet rule win over a tw utility

kane50613:
## Why

The `tw` prop was merged above every matched stylesheet rule, so a utility beat any selector regardless of specificity: `#hero { background: blue }` lost to `tw="bg-red-500"`. Tailwind puts its utilities in `@layer utilities`, which unlayered CSS outranks, so the engine had the relationship backwards.

## What

Utilities now merge below every author rule, and a utility marked `!` merges above them, because the cascade reverses layer order for important declarations. `StyleDeclarationBlock::split_importance` separates the two halves of a `tw` block so each lands at its own end.

The `tw_cascade` fixture pins both halves. Simulating the old order turns its first two boxes red; the new order paints the stylesheet's blue and keeps red only where the utility says `!`. One case is still open and deliberately left out of the fixture: an important utility against an important stylesheet rule computes the utility's colour in `ComputedStyle` but paints the stylesheet's, in both the raster and SVG backends. That points at a second style path rather than at this ordering, so it wants its own investigation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CSS precedence so standard Tailwind utilities are overridden by matching stylesheet rules.
  * Ensured `!` Tailwind utilities take precedence over important stylesheet rules.
  * Preserved declaration ordering and cascade behavior across supported styling scenarios.

* **Documentation**
  * Added guidance on Tailwind utility precedence and recommended approaches for existing overrides.


kane50613:

PR #1391, opened 2026-08-24
fix(helpers): enforce fetch policy on cached images

### LINKED ISSUE #1390: fetchCache bypasses the allowUrl and maxBytes fetch policies
## Summary

`fetchImageData` looks up `fetchCache` before it applies any fetch policy. A cache hit returns bytes without running the current call's `allowUrl` and without checking its `maxBytes`.

https://github.com/kane50613/takumi/blob/master/takumi-helpers/src/utils.ts#L186-L192

```ts
const cached = fetchCache?.get(url);
if (cached) {
  return cached;
}
```

`allowUrl` runs inside `fetchOk`. `maxBytes` runs inside `readBodyLimited`, after the fetch. Both run after the cache lookup, so a cached entry skips them.

## Impact

The docs define `allowUrl` as an SSRF allowlist and enforce it on every redirect hop. The same docs support sharing one `fetchCache` across renders. A service that shares a cache between tenants, routes, or security levels loses policy isolation:

1. Request A uses a permissive policy and caches 5 MB for a URL.
2. Request B passes `allowUrl: () => false`, or `maxBytes: 1 MB`.
3. Request B gets the cached 5 MB. Its policy never runs.

Existing tests cover `allowUrl`, redirects, `maxBytes`, and cache single-flight, but no test covers the cache + policy combination.

## Suggested fix

Enforce both policies on cache hits in `fetchImageData`:

- Run the current call's `allowUrl(url)` before returning a cached entry.
- Re-check `data.byteLength` against the current call's `maxBytes`.

Two constraints keep behavior stable:

- Keep both checks on the cache-hit branch only. The miss path must keep calling `allowUrl` exactly once per hop inside `fetchOk`; a pre-lookup check would call it twice.
- Never evict on a policy rejection. The entry stays valid for callers with other policies; eviction stays reserved for fetch failures.

luojiyin1987:
Fixes #1390

## Problem

`fetchImageData` returned a `fetchCache` hit before running any policy. `allowUrl` (SSRF allowlist, enforced on every redirect hop) and `maxBytes` only ran on the fetch path. A cache shared across tenants, routes, or security levels let one call resolve bytes that another call's policy forbids.

Review of the first cut surfaced three deeper gaps in the same area:

1. Single-flight stores one promise, so a consumer with a looser `maxBytes` inherited the creator's tighter rejection.
2. A cache hit checked only the entry URL. Bytes fetched through a redirect into a blocked host still passed a strict hostname policy.
3. An in-flight entry's recorded redirect chain grows while the shared promise runs. Validating it before fulfillment missed redirects discovered later.

## Change

Entries now carry metadata recorded at creation time (the creator's byte cap and every URL its `allowUrl` checked, entry plus redirect hops). The metadata lives in a module-level `WeakMap` keyed by the entry promise, so `ImageFetchCache` values stay plain promises and the public interface is unchanged.

On a cache hit, the current call's policy runs in full:

- For trusted metadata, `allowUrl` runs once after the shared promise fulfills. At that point the redirect chain is complete. This blocks redirects discovered after an in-flight cache hit.
- Resolved bytes re-check against the current call's `maxBytes`.
- If the shared stream rejects under its creator's tighter cap while this call allows more, this call refetches privately instead of failing too; unrelated failures keep propagating to every consumer.
- An entry with no recorded chain — created without `allowUrl`, pre-seeded by hand, or produced by another module instance — is unprovenance. A caller carrying `allowUrl` does not trust it and refetches privately, because a miss under that policy would have checked every hop. Fail-closed: metadata present means revalidate after fulfillment, no metadata means no reuse.

Supporting internals:

- `readBodyLimited` now throws an internal `ResponseTooLargeError` carrying its cap, so cap-mismatch detection between cache consumers is type-based instead of message-text matching. The class is not exported; error messages are unchanged.

Constraints kept from earlier cuts:

- The miss path is untouched, so the network path still calls `allowUrl` exactly once per hop inside `fetchOk`.
- Policy rejections never evict the entry. Eviction stays fetch-failure-only.

## Tests

Regression tests in `takumi-helpers/tests/fetch-limits.test.ts`:

- a cache hit still respects a stricter `allowUrl` policy
- a cache hit rechecks every recorded redirect hop
- an in-flight cache hit rechecks redirects discovered later
- a strict caller does not trust cache created without `allowUrl`
- a stricter maxBytes rejection keeps the shared entry reusable
- concurrent cache consumers keep independent `maxBytes` limits
- a looser consumer reuses cached bytes that fit

Also adds `.tegami/fetch-cache-policy.md` (patch, `@takumi-rs/helpers`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image caching so cached results are revalidated against URL access rules and byte-size limits.
  * Prevented cached images from bypassing redirect and security policies.
  * Automatically refetches images when cached data lacks required policy information or exceeds the current request’s limits.
  * Preserved independent limits for concurrent image requests.
  * Ensured oversized image responses are handled consistently without invalidating reusable shared cache entries.

### DIFF
--- a/.tegami/fetch-cache-policy.md
+++ b/.tegami/fetch-cache-policy.md
@@ -6,4 +6,4 @@ packages:
 
 ### Enforce fetch policies on shared image cache entries
 
-`prepareImages` with a shared `fetchCache` returned cached bytes without running the calling render's `allowUrl` or `maxBytes`, so bytes fetched under one tenant's policy could serve a render whose policy forbids them. Cache hits now run `allowUrl` over the URL and its recorded redirect chain, re-check `maxBytes` against the resolved bytes, and refetch privately when the shared entry was capped tighter than the current call allows.
+`prepareImages` with a shared `fetchCache` returned cached bytes without running the calling render's `allowUrl` or `maxBytes`, so bytes fetched under one tenant's policy could serve a render whose policy forbids them. Cache hits now run `allowUrl` over the URL and its recorded redirect chain, re-check `maxBytes` against the resolved bytes, and refetch privately when the shared entry was capped tighter than the current call allows. Entries with no recorded redirect chain are not trusted by callers carrying `allowUrl`; those callers refetch them under their own policy.
--- a/.tegami/fetch-cache-policy.md
+++ b/.tegami/fetch-cache-policy.md
@@ -1,0 +1,9 @@
+---
+packages:
+  "@takumi-rs/helpers":
+    type: patch
+---
+
+### Enforce fetch policies on shared image cache entries
+
+`prepareImages` with a shared `fetchCache` returned cached bytes without running the calling render's `allowUrl` or `maxBytes`, so bytes fetched under one tenant's policy could serve a render whose policy forbids them. Cache hits now run `allowUrl` over the URL and its recorded redirect chain, re-check `maxBytes` against the resolved bytes, and refetch privately when the shared entry was capped tighter than the current call allows.
--- a/takumi-helpers/src/utils.ts
+++ b/takumi-helpers/src/utils.ts
@@ -170,7 +170,8 @@ function extractImageUrls(node: Node): string[] {
  * requests for the same URL (single-flight) and reuses their bytes. Any object with `Map`-like
  * `get`/`set`/`delete` works, so LRU/TTL policies can be plugged in. Each entry remembers the
  * options of the call that created it, so later calls still run their own `allowUrl` and
- * `maxBytes` against the shared bytes.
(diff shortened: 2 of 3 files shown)


kane50613:
Thanks for digging into #1390 — the redirect-chain recheck idea here was right, and #1396 keeps it. We went with a smaller fix overall: cache hits re-run the caller's allowUrl (over the recorded chain) and maxBytes, and entries fetched without allowUrl fall back to an entry-URL check, documented as a limitation. The typed size error, private refetch path, and provenance trust model added more surface than the issue needs, and the private refetch bypasses single-flight for policy-carrying callers. Closing in favor of #1396.

PR #1437, opened 2026-08-27
Bindings: warn when a render takes the deprecated stylesheets option

kane50613:
## Why
`takumi-js` and `takumi-pdf` warn once when a caller passes `stylesheets`, but `@takumi-rs/core` and `@takumi-rs/wasm` took the alias in silence. Anyone calling the bindings directly gets no signal before the alias is removed.

## What
`resolve_css` in `takumi-bindings-common` picks `css` over the alias and hands the deprecation message to a warn callback the first time the alias appears. napi reaches `console.warn` through the `Env` its task constructors already receive, and wasm through a `wasm-bindgen` extern, so neither binding gains a dependency.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSS configuration handling across rendering, animation, SVG, and measurement.
  * The deprecated `stylesheets` option remains supported as an alias for `css`.
  * A deprecation warning now appears in the console only once, when the alias is first used.
  * When both options are provided, the standard `css` option is used consistently.
* **Documentation**
  * Added release notes for the updated bindings and PDF package.

### DIFF
--- a/.tegami/bindings-stylesheets-warning.md
+++ b/.tegami/bindings-stylesheets-warning.md
@@ -1,0 +1,13 @@
+---
+packages:
+  "@takumi-rs/core":
+    type: patch
+  "@takumi-rs/wasm":
+    type: patch
+  "takumi-pdf":
+    type: patch
+---
+
+### Warn when the bindings take the deprecated `stylesheets` option
+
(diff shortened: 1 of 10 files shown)


kane50613:
Dropping this. The wrapper packages already warn, and the binding option carries an @deprecated tag that editors surface, so reaching into the host console from Rust bought nothing.

PR #1439, opened 2026-08-27
Render an unknown counter style as decimal

kane50613:
`list-style-type` rejected any counter style name takumi does not implement, so `list-style-type: georgian` dropped the whole declaration and the item kept whatever marker it inherited.

A `<counter-style-name>` is a `<custom-ident>`, so those names parse. Markers for a name with no definition fall back to `decimal`, matching Blink (`counter_style.cc:908`).

Idents that css-values excludes from `<custom-ident>` still fail: the CSS-wide keywords, `revert`, `revert-layer`, and `default`.

The Tailwind `list-*` prefix now tries `list-style-position` before `list-style-type`, the same order `ListStyleShorthand` uses. Without that, `list-inside` would parse as a counter style name.

`list_markers.html` gains a cell for the fallback.

## Summary by CodeRabbit

* **New Features**
  * List markers now accept any counter style name.
  * Unsupported counter styles automatically fall back to decimal numbering, matching browser behavior.
  * Added coverage for unknown counter styles in rendered lists.

* **Bug Fixes**
  * Improved handling of list-style declarations with unrecognized counter style names.


kane50613:
Closing. `style={{}}` is error-first, and this change breaks that: `listStyleType: "zzz-nope"` went from a throw naming the valid values to a silent decimal marker. Browser-style fallback makes sense for a stylesheet, which is parsed loosely anyway, but not for a hand-written style object. Splitting the behaviour by source would be worse than the gap it closes.

PR #1464, opened 2026-08-28
PDF: name takumi in the document's Producer

kane50613:
## Why

A takumi PDF had no `/Producer`, no `/Creator` and no `xmp:CreatorTool`, so nothing in the file named the renderer. The traces it did leave came from krilla, which typst also uses.

## What

Every document writes `takumi-pdf <version>` as `/Producer` and `pdf:Producer`, from the serializer rather than from the metadata object, so a document that sets no metadata carries it too. The crate carries a version for the first time and joins the npm package in a tegami group so the two stay equal.

PDF 2.0 deprecates `/Producer`, so a PDF/A-4 document gets the XMP entry only. The npm and wasm surfaces cannot change the value.

## Summary by CodeRabbit

* **New Features**
  * Rendered PDFs now identify `takumi-pdf` and its version in PDF producer metadata.
  * Added support for customizing the producer value through PDF metadata options.
  * Producer information is included in both the PDF information dictionary and XMP metadata.
* **Bug Fixes**
  * Ensured producer metadata is consistently written, including PDFs without other document metadata.
* **Chores**
  * Updated the `takumi-pdf` package version to 0.11.3.


kane50613:
Reworked: the producer now comes from the serializer instead of `PdfMetadata`, so it lands on every document including PDF/A ones. The CodeRabbit note about PDF/A documents omitting it no longer applies, and the `metadata.producer` option it referenced is gone from the npm and wasm surfaces.

PR #1468, opened 2026-08-29
WebP: encode only each animated frame's changed region

kane50613:
## Why

Every frame of an animated WebP was encoded as a full canvas, so an animation whose motion covers a small part of the frame paid for the whole frame on every one. The only saving in place was merging consecutive byte-identical frames into one longer frame.

## What

Frames after the first now carry only the bounding box of the pixels that changed, written as an ANMF sub-rectangle with the no-blend flag so the rectangle replaces the canvas rather than compositing against it, which is what keeps partial alpha correct. A sweeping-highlight animation (544×216, 42 frames, lossless) drops from 61,512 to 22,802 bytes and from 160 ms to 40 ms; a frame where everything changes is byte-identical to before. Cropping is skipped for `dispose: true`, where the canvas no longer holds the previous frame.

Two pre-existing problems surfaced while measuring and are left alone here: animated lossy frames with alpha never carry an `ALPH` chunk, and merged durations are clamped to 24 bits only at write time, so a hold longer than 0xffffff ms silently loses time.

## Summary by CodeRabbit

* **New Features**
  * Animated WebP encoding now stores only changed regions for subsequent frames when possible.
  * Improved animation output with smaller file sizes and faster encoding for localized motion.
  * Frame placement and blending are handled automatically, including transparent pixels and disposal behavior.

* **Bug Fixes**
  * Corrected animated WebP frame regions and replacement behavior for accurate playback.


kane50613:
Skipping the flag change, adding the test it asked for.

Honouring `options.blend` on a full-canvas diff region would break the invariant the whole diff pass rests on. Each frame is compared against the previous *source* frame, which only works while the canvas equals that frame. A full-canvas partial-alpha frame emitted with blend composes the two alphas, so the canvas stops matching the source, and every later cropped frame under-covers the real difference: stale blended pixels survive outside the rectangle. Uniform no-blend in the diff path is what keeps that induction sound, and it makes the composited output match the frames the caller passed in, which the old blend path did not.

The regression test is in `fc3ec4b24`: `full_canvas_partial_alpha_frames_replace_under_blend` encodes four full-canvas alpha-128 frames with `blend: true` and asserts every composited frame is bit-identical to its source. It fails under the proposed change.

PR #1474, opened 2026-08-29
Raster: zero scratch buffers instead of leaving them uninitialized

kane50613:
## Why

Every raster scratch buffer came from `uninit_buffer`, which called `set_len` on
a `Vec` it had only reserved. That is undefined behavior the moment it returns,
and it is what turns a wrong length into an information leak: a caller that
fills only part of the buffer paints uninitialized heap bytes into the output.
The base PR fixes two callers that did exactly that.

## What

`uninit_buffer` is gone and every call site allocates `vec![0; n]`, so a
partially written buffer paints black instead of heap contents. `checked_area`
moves next to the pixmap budget it belongs with, which leaves `canvas/scratch.rs`
empty and deleted.

Fixture goldens are byte-identical. An A/B over the `effects`, `canvas` and
`background` suites found no cost outside this machine's noise floor, which a
zero-change control run showed reaching 43%.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved rendering reliability for large images, blur effects, shadows, filters, and text masks.
  - Added safeguards against buffer-size overflows and invalid or oversized rendering areas.
  - Improved handling of malformed alpha-channel data to prevent unexpected rendering failures.

- **Security & Stability**
  - Rendering buffers are now safely initialized, reducing corrupted or unpredictable visual output.
  - Improved stability when processing empty, oversized, or overflowing image areas.


kane50613:
Folded into #1473 — one PR now carries the crash fixes and the uninit-buffer removal.

PR #1481, opened 2026-08-30
Playground: let templates import echarts

kane50613:
## Why
Templates can only import `takumi-pdf/primitives`, so one that wants a chart has nowhere to get it.

## What
Templates can now import the `echarts/core` entry points, fetched lazily as their own chunks so the worker bundle stays the same size, with typings wired into the editor and an Analytics chart template that passes the SVG straight to an `img` src. The editor typings also moved to lazy chunks, which shrinks the playground chunk from 1.8 MB to 265 kB, and a shared link's output pane now asks the reader to review the code instead of showing a spinner.

Analytics chart template rendered in the playground

Shared link output pane prompting to read the code before Run

## Summary by CodeRabbit

* **New Features**
  * Added an analytics chart template for PNG dashboard visuals.
  * Added ECharts support for chart-based playground templates, including analytics and rendering reports.
  * Reports now include latency, render-volume charts, totals, and visualization legends.
  * Added clearer guidance when shared playground code has not yet been run.

* **Bug Fixes**
  * Improved inline SVG rendering, chart code completion, and module loading.
  * Fixed stale playground output when opening shared links or clearing code.
  * Enhanced compatibility when rendering chart-based templates.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -27,6 +27,7 @@
         "beautiful-mermaid": "^1.1.3",
         "class-variance-authority": "^0.7.1",
         "clsx": "^2.1.1",
+        "echarts": "^6.1.0",
         "fumadocs-core": "^16.14.5",
         "fumadocs-mdx": "^15.3.0",
         "fumadocs-twoslash": "^3.3.0",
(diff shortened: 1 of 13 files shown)


kane50613:
The Paged table PDF template now opens with a latency line chart and a weekly renders-by-format stacked bar chart above the table, both rendered by echarts into raw SVG image sources.

Two-chart PDF report

Note for local dev: the missing chart axis labels turned out to be a stale `takumi-pdf-js/pkg` wasm build, not a rendering bug. Rebuild with `bun run build` in `takumi-pdf-js` and SVG `<text>` embeds fine.

PR #1482, opened 2026-08-30
Core: apply text-fit under letter and word spacing

kane50613:
## Why
Any `letter-spacing` or `word-spacing` on a text span silently disabled `text-fit`, so a spaced nowrap headline overflowed its container instead of shrinking.

## What
Spacing no longer turns the fit off: the line scales uniformly about its origin, so the fitted line width matches Blink's `text_fit_utils.cc` result while the spacing scales with the glyphs instead of staying constant, a drift the code comment names. Floats still disable the fit, and the letter-spacing and word-spacing shrink/grow combinations now have measure tests plus a golden fixture.

## Summary by CodeRabbit

* **New Features**
  * Text fitting now works with headlines that use letter spacing or word spacing.
  * Spaced text can automatically grow or shrink to fit its available space.
  * Supports more consistent headline sizing across different layouts.

* **Bug Fixes**
  * Fixed spaced headlines overflowing because text fitting was silently disabled.
  * Improved rendering bounds for positioned floating content.

* **Tests**
  * Added coverage for growing and shrinking text with letter and word spacing.


kane50613:
The new golden: both spaced headlines used to overflow their white boxes; they now shrink to fit.

text-fit with spacing golden

PR #1504, opened 2026-09-01
Core: break lines where Chromium breaks them

kane50613:
## Why

Line breaking followed ICU's defaults, so text wrapped at points a browser would not pick. parley 0.11 ships Chromium's own break table, and takumi never enabled it.

## What

Inline layout hands `CHROMIUM_LINE_BREAK_OVERRIDE` to the parley builder. The override is paragraph-wide and replaces every opportunity the segmenter found, so a paragraph containing a `word-break: break-all` span skips it; Blink runs break-all through a separate iterator anyway.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated text wrapping to follow Chromium-compatible line-breaking behavior.
  - Hyphenated text such as `ISBN-2026408` now wraps at the hyphen, matching browser rendering.
  - Preserved `word-break: break-all` behavior, allowing breaks between every character.
  - Improved consistency for text wrapping in regular content and truncated text with ellipses.

- **Tests**
  - Added coverage for standard and `break-all` text wrapping scenarios.


kane50613:
Verified against parley 0.11.1: a per-span override is not expressible. `set_line_break_override` lives on the builder (`builder.rs:58`), `StyleProperty` has no override variant, and `LineBreakContext` carries only three code points, so a closure cannot tell which span a break position belongs to. The paragraph-wide scope is documented on `chromium_line_breaks`, and the fixture's second column pins that break-all still breaks per character. The fix belongs upstream, where `analyze_text` already iterates per word-break style run.

PR #1505, opened 2026-09-01
Core: accept media query range syntax

kane50613:
## Why

`@media` only parsed the `min-width: 768px` form, so a stylesheet written with the Media Queries Level 4 range syntax failed to parse and took the whole sheet down with it.

## What

`(width >= 768px)`, the reversed `(768px <= width)`, and the two-sided `(400px < height <= 700px)` all parse now, for `width` and `height`, with `calc()` values and the boolean `(width)` form. A range whose two comparisons point opposite ways is rejected, matching how a browser drops the query.

Still unsupported and out of scope here: `or` and nested query groups, and features beyond `width`, `height`, and `orientation`.

## Summary by CodeRabbit

* **New Features**
  * Added support for modern media query range syntax, including width and height comparisons.
  * Supports chained comparisons, reversed operands, units, and `calc()` expressions.
  * Bare feature queries, such as `(width)`, now match non-zero viewport values.
  * Media feature names and prefixes are now matched without regard to letter case.

* **Bug Fixes**
  * Rejects invalid media queries with opposing range comparisons.

* **Tests**
  * Added coverage for range-based media queries across viewport sizes and dimensions.


kane50613:
Fixed: the base feature name is now taken with an ASCII-case-insensitive prefix match, so `(MIN-WIDTH: 600px)` resolves to `width` with the `Min` comparison. Mixed-case cases added to the tests.

PR #1507, opened 2026-09-01
Core: apply @media print rules to PDF output

kane50613:
## Why

`@media print` never matched, so a stylesheet written for print had no effect on a PDF render, and `@media screen` applied to output that is not a screen.

## What

PDF renders resolve the `print` media type and image renders resolve `screen`, carried by the new `Viewport::media_target`. Every viewport takumi-pdf lays out against sets it, including the one a caller passes in.

## Summary by CodeRabbit

* **New Features**
  * PDF rendering now supports `@media print` styles.
  * Image rendering continues to use `@media screen` styles.
  * Viewports can be configured for screen or print media targets.

* **Bug Fixes**
  * Improved media-query handling for print, screen, combined, and negated conditions.
  * PDF output now correctly ignores screen-only styles and applies print-specific styles.

* **Documentation**
  * Added guidance on media targets used by PDF and image rendering.


kane50613:
The changeset marks `takumi-core` as `minor`, which takes 0.23.1 to 0.24.0 on release. Keeping the state outside `Viewport` was the alternative considered: both call sites that evaluate media queries (`matching.rs` and `layout/tree.rs`) receive only a `Viewport`, so anywhere else would have to be threaded through them.

PR #1508, opened 2026-09-01
Core: match resolution and aspect-ratio media queries

kane50613:
## Why

`resolution` and `aspect-ratio` were unsupported, though both follow from data the renderer already holds: the device pixel ratio and the viewport size.

## What

`@media (resolution >= 2dppx)` reads the render's device pixel ratio and accepts `dppx`, `x`, `dpi`, and `dpcm`; `@media (aspect-ratio >= 16/9)` reads the viewport ratio. A media feature value is now parsed before its name is known, which is what the reversed range form `(2dppx <= resolution)` needs.

## Summary by CodeRabbit

* **New Features**
  * Added support for CSS resolution media queries using `dppx`, `x`, `dpi`, and `dpcm`.
  * Added aspect-ratio media queries based on viewport proportions, including equality, range, reversed-range, and unit-conversion comparisons.
  * Added support for range syntax and `min-`/`max-` prefixes across supported media features.
  * Improved matching precision for resolution and viewport dimensions, including inclusive boundaries and negative values.

* **Tests**
  * Added automated coverage and visual fixtures for resolution and aspect-ratio media queries.


kane50613:
All three fixed:

- A zero or negative ratio divisor is rejected, so `(aspect-ratio <= 1/0)` becomes `not all` instead of matching everything.
- The 0.5 tolerance is now passed per feature: pixels keep it, `resolution` and `aspect-ratio` use 1e-5. `(resolution: 1dppx)` no longer matches a 1.49 device pixel ratio.
- `MediaFeature::boolean` builds the value kind each feature takes, so `(resolution)` and `(aspect-ratio)` evaluate instead of parsing as unsupported.

PR #1534, opened 2026-09-02
PDF, HTML: render form controls as fillable fields

kane50613:
`form: true` emits `<input>`, `<textarea>` and `<select>` as AcroForm fields, so one template covers both the fillable form and the printed copy. Off by default, so no existing golden moves.

## What it maps

The HTML attributes carry the field. There is no new component vocabulary.

| HTML | PDF |
| --- | --- |
| `name` | Field name. Radio buttons sharing one become a group with `/Kids`. |
| `value`, `checked` | `/V` and `/DV`. |
| `required`, `readonly`, `disabled` | `/Ff`. |
| `maxlength` | `/MaxLen`. |
| `<textarea>`, `<input type="password">` | Multiline and password flags. |
| `<select>`, `<option value>` | `/Ch` with `/Opt` display and export pairs. |
| `aria-label`, `<label for>`, `title`, `placeholder` | `/TU`. |
| `color`, `font-size`, `background-color`, `border-color`, `text-align` | `/DA`, `/MK`, `/Q`. |

The control paints through the normal CSS pipeline and the widget's appearance stream draws only the value on top, so border, background and radius are whatever the stylesheet says. `/MK` and `/DA` carry the same colors so a reader redrawing an edited field lands close to the rendered box.

## Correctness details worth reviewing

- A radio button's appearance state is its index in the group, and the readable value lives in the group's `/Opt`. A value like `Credit Card` would not survive as a PDF name.
- A check box takes its on-state from `value`, escaped, defaulting to `on` like a browser submit.
- Form controls are unbreakable atoms in pagination. A widget has one rectangle on one page, so a control cut in half would lose its field.
- Widgets carry `/P`, so a reader does not scan every page's `/Annots` to find a field's page.
- Tagged output wraps each widget in a `Form` structure element, and everything the control paints becomes an artifact, because PDF/UA admits one child there.
- Two controls sharing a `name` without being one radio group fail the render. Merging them into one field would show the same value in both boxes.

## Behavior changes outside the option

Form controls now lay out as block-level boxes. An inline-level box has no rectangle of its own for a widget to cover, and an `<input>` inside a paragraph previously painted nothing at all. `<option>` no longer paints its text into the page, matching how a `<select>` shows only its selection.

## Known limits

Values draw with the standard Helvetica face, so a prefilled value outside the Latin alphabet does not render, and a document with prefilled values fails PDF/UA on font embedding. A blank form embeds no such font and passes veraPDF UA-1. Embedding a real face for `/DR` and the appearance streams is the follow-up.

Submit and reset actions, comb fields and reading values back out are not included.

## Testing

169 `takumi-pdf` tests and 25 `takumi-pdf-js` tests pass, clippy is clean on the touched crates, and no existing golden changed. New golden: `takumi-pdf/tests/fixtures-generated/form_fields.pdf`. veraPDF UA-1 on a filled form reports only the font embedding rule.

## Summary by CodeRabbit

* **New Features**
  * Added support for generating fillable PDF forms from HTML `<input>`, `<textarea>`, and `<select>` elements when enabled.
  * Supports text, multiline, password, checkbox, radio-button, and drop-down fields.
  * Added options for configuring the font used by form fields.
  * Form controls preserve labels, accessibility attributes, styling, and appropriate PDF structure.

* **Bug Fixes**
  * Prevented form controls from splitting across pages.
  * Rendering reports errors for conflicting field names outside radio groups.

* **Documentation**
  * Added guidance for creating and configuring fillable PDF forms.


kane50613:
Superseded by the stack #1536 → #1537 → #1538 → #1539, which splits this into one reviewable layer per concern and fixes the 21 findings from the CodeRabbit and Codex audits on this PR.

PR #1546, opened 2026-09-05
Helpers: recover from transient font and image requests

kane50613:
## Why

A temporary font-service failure rejects the render immediately. Retrying at the call site duplicates policy and can extend the request deadline.

## What

Retry transient GET and HEAD failures within the existing timeout, including font downloads and build-time Google Fonts metadata. Preserve cancellation, redirect checks, permanent errors, and server retry delays without adding options.

## Summary by CodeRabbit

* **Bug Fixes**
  * Font and image requests now automatically retry transient network failures, timeouts, rate limits, and temporary server errors.
  * Google Fonts metadata and font downloads are more resilient to temporary service interruptions.
  * Redirected requests retain their existing security and policy checks.
  * Permanent errors and non-retryable request types are not retried.

* **Documentation**
  * Timeout behavior now reflects the total time allowed for the request, including retries.

### DIFF
--- a/.tegami/font-fetch-retries.md
+++ b/.tegami/font-fetch-retries.md
@@ -1,0 +1,9 @@
+---
+packages:
+  "@takumi-rs/helpers":
+    type: patch
+---
+
+### Retry transient font and image requests
+
+Retry temporary GET and HEAD failures within the existing timeout, including Google Fonts metadata and font file downloads.
--- a/takumi-helpers/src/utils.ts
+++ b/takumi-helpers/src/utils.ts
@@ -3,6 +3,9 @@ import type { Node } from "./types";
 
 const defaultFetchTimeout = 30_000;
 const maxRedirectHops = 5;
+const maxFetchAttempts = 3;
+const maxRetryDelay = 1000;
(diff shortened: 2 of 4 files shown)


kane50613:
Measured the Google Fonts download stage locally with actual CSS and WOFF2 responses, including the renderer's codepoint filtering and list-marker characters:

| Input | CSS subsets | Selected downloads | Bytes |
| --- | ---: | ---: | ---: |
| Short Chinese text, Noto Sans TC | 105 | 9 | 257,980 |
| Chinese paragraph, Noto Sans TC | 105 | 11 | 336,116 |
| Short Chinese text, TC weights 400/500/700 and JP weights 400/700 | 229 | 28 | 1,054,376 |

All measured downloads succeeded. Alternating unrestricted downloads and a six-request limit did not show a consistent latency benefit from limiting concurrency. For the 28-file case, unrestricted runs took 506/104/78 ms and limited runs took 246/142/128 ms. These are local download timings, not CI or end-to-end render timings.

No concurrency limit was added. Multiple CJK families can increase request count, but this experiment did not reproduce the reported CI timeout or establish concurrency as its cause. The docs now distinguish the CSS request from lazy font downloads.

PR #1595, opened 2026-09-10
Docs: highlight the playground with sugar-high

kane50613:
## Why

The playground editor loaded shiki's oniguruma wasm engine and the TSX TextMate grammar only to colour Monaco, and both had to arrive before the first token was painted. sugar-high tokenizes the same TSX in plain JavaScript, so the route drops roughly 275 KB gzipped (890 KB raw) without changing what the editor does.

## What

- Added a Monaco tokens provider backed by sugar-high: it parses the whole model once per version, caches that per model, and serves the per-line tokens Monaco asks for.
- Defined `takumi-light` and `takumi-dark` Monaco themes with GitHub's default colours in place of the `github-*-default` themes shiki registered, and dropped `@shikijs/monaco` from `docs`. The home page and MDX still use shiki, so `shiki` itself stays.
- The editor mounts before its type definitions load: React, csstype, takumi and the playground options are registered as extra libs right after `onMount`, and the megabytes of echarts typings wait until the code mentions `echarts`.
- The tokenizer state carries the token left open at the end of a line instead of the model version, so Monaco stops re-tokenizing at the first line whose context is unchanged: editing a line in the middle of the welcome template now costs one `tokenize` call instead of 23.
- The provider looks each line up in the parsed models and falls back to the last matched model when more than one model holds that line, which the playground's single model makes exact.
- Both themes are now derived from `@shikijs/themes`' `github-dark-default` and `github-light-default` at registration time, mapping each sugar-high token to the TextMate scope GitHub colours it with, which adds about 30 KB raw to the playground chunk.

Chunk sizes from `waku build`, raw / gzip:

| chunk | before | after |
| --- | --- | --- |
| `wasm-*.js` (oniguruma) | 622,325 / 231,160 | gone |
| `tsx-*.js` (grammar) | 175,529 / 16,597 | gone |
| `github-dark-default`, `github-light-default`, hast helpers | 47,481 / 12,843 | gone |
| `playground-*.js` | 274,587 / 86,585 | 189,925 / 61,100 |
| `client-*.js` | 210,447 / 65,686 | 228,899 / 70,624 |

Client JS across the whole site: 6,842,347 → 5,930,786 bytes.

Known differences from the shiki grammar: sugar-high has no function token, so a call like `slug(...)` keeps the identifier colour instead of GitHub's purple, and a member access such as `.replace` takes the same blue as a JSX attribute name. Multi-line comments, template literals, regex literals, and JSX all tokenize correctly.

## Summary by CodeRabbit

* **Improvements**
  * Updated the playground code editor with locally integrated syntax highlighting and light/dark themes.
  * Improved highlighting for numeric constants and code following multi-line comments or template literals.
  * Improved editor responsiveness by loading code typings incrementally.
  * Added more reliable recovery when typings fail to load, allowing later attempts to succeed.

* **Bug Fixes**
  * Improved syntax updates when editing code, keeping highlighting accurate throughout the file.


kane50613:
Closing: the playground keeps shiki for highlighting. The typings-after-mount part moved to #1617, and #1600 / #1601 now build on it.

PR #1603, opened 2026-09-10
Docs: move the playground editor to CodeMirror

kane50613:
## Why

Monaco reached the browser as roughly 2.6 MB of gzipped JavaScript from a CDN before the first keystroke, and none of it could be tree-shaken. Its TextMate highlighting needed the oniguruma wasm on the main thread, and the alternative, coarser highlighters give up on TypeScript.

## What

The playground now runs on CodeMirror 6, with the same features it had and the same colours, and the TypeScript language service moved into a worker that starts once the editor has painted.

- Editor: line numbers, folding, bracket matching and closing, undo, wrapping, an active line highlight, autocompletion, multiple cursors, ⌘↵ to run, ⇧⌥F to format, and the same phone layout without line numbers. The keymap is VS Code's, and ⌘F opens a find and replace widget laid out like VS Code's, with a match count and case, word and regexp toggles.
- Themes: GitHub Dark Default and GitHub Light Default are derived from the shiki theme data, so `colors` drives the chrome and `tokenColors` drives a `HighlightStyle` resolved by the most specific dotted scope.
- Colours: a small syntax-tree pass adds the scopes VS Code's grammar assigns from context, such as `const` names, primitive types, JSX braces and operators, so every template matches shiki token for token apart from a few TextMate quirks.
- TypeScript: the same typings load into a `@typescript/vfs` map inside the worker, with echarts pulled in only once the snippet mentions it. Autocompletion, hover, diagnostics and signature help all come from that one worker, shared by both breakpoints' editors.
- Size: the playground chunk grows from 84.4 KB to 219.0 KB gzipped, which replaces 2.6 MB of CDN Monaco. The 1.36 MB gzipped TypeScript worker is fetched after the first paint, and the typings chunks are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Replaced the playground’s Monaco editor with a CodeMirror-based editor.
  * Added TypeScript linting, autocomplete, hover information, and signature help.
  * Added lazy loading for TypeScript and ECharts typings as needed.
  * Added improved syntax highlighting, theme support, and code wrapping.
  * Added find-and-replace with search navigation, match counts, and replacement controls.
  * Added keyboard shortcuts for running and formatting code.
  * Improved editor responsiveness, selection behavior, tooltips, and fold controls.


kane50613:
Closing: the playground stays on Monaco. The local bundle (#1600) and deferred TypeScript service (#1601) cover the loading cost instead.

PR #1625, opened 2026-09-13
CSS: read @page size and margin into the page options

kane50613:
## Why
`@page { size; margin }` is how a Chromium print stylesheet sets its page geometry, and the Puppeteer migration table listed it as unsupported.

## What
- The stylesheet parses `@page` into `PageRule`: a selector list (a page name plus `:first`, `:left`, `:right`, `:blank`, kept for named pages), a `size` of `auto`, lengths, a page keyword and an orientation, and the margin descriptors with their `!important`. `@media` and `@layer` apply to it like any rule.
- In paged output the descriptors that win the cascade for every page lay over the page options: `size` replaces the page size and each margin descriptor its side, resolved against that size, so CSS wins over the options the way `preferCSSPageSize` does. A rule with a page selector is parsed and left alone until named pages apply it.

Stacked on #1623. The `at-page-size-and-margin` golden renders under `@page { size: A5 landscape; margin: 10mm 5mm }` beside a `@page cover` rule and a `@media screen` rule that must not apply.

## Summary by CodeRabbit

- **New Features**
  - Added support for CSS `@page` rules in PDF generation.
  - Page size, orientation, and margins can now be defined in CSS and take precedence over render options.
  - Supports named and custom page sizes, landscape orientation, media conditions, cascading rules, and percentage margins.

- **Documentation**
  - Updated PDF documentation with `@page` usage examples and Puppeteer option mappings.
  - Documented currently unsupported page-selector variants.

- **Tests**
  - Added coverage for page sizing, margins, cascading, and media conditions.


kane50613:
Closing in favour of the option-level page bands in #1627. The `@page` parser and named-page contexts stay in this branch's history if a CSS-driven page geometry comes back.

PR #1626, opened 2026-09-13
PDF: style named pages with page and @page <name>

### LINKED ISSUE #1571: Allow a page to opt out of document margin and header/footer bands (full-bleed cover)
**What happens:** `margin` and header/footer bands are document-wide. An absolutely positioned
full-page image with negative offsets is clipped to the content box, and the footer band prints on
the cover too. The workaround (`margin: 0`, every page as its own fixed-height box of
`pageHeight − 1px`, footer drawn manually per page) works but gives up automatic pagination for
flowing content and needs the `−1px` to avoid a rounding overflow onto the next page.

**Ask:** something like `style={{ breakBefore: "page", pageMargin: 0, pageBands: "none" }}` on a
page-level element, or a `pages: [{ margin, header, footer }]` override keyed by page index.
Also: `pageHeight`-tall boxes overflow their page by a sub-pixel; a documented "exactly one page"
helper (or clamping) would remove the `−1px` folklore.

**Context:** gradients render as real PDF shadings with SMask
alpha, SVG data-URI images embed as vectors, `PageNumber`/`TotalPages` resolve outside footer bands,
3 pages with two JPEGs render in ~40 ms. That is why we are moving our production booklet onto Takumi.

_Tested with takumi-pdf 0.14.1 / @takumi-rs/helpers 2.13.5 on macOS, Node 22; latest 0.14.3 / 2.13.7 release notes show no font-related changes._

kane50613:
Fixes #1571

## Why
A full-bleed cover needs its own page margins, and the only way to get them was to give up automatic pagination and draw every page by hand. CSS names pages for this: `page: cover` on the element and `@page cover { margin: 0 }`.

## What
- `page` is a new inherited property, and a run of adjacent siblings naming the same page opens a page group that the matching `@page <name>` rule sizes and margins. The group is wrapped in a page-area box laid out in the same column at that page's content width, shifted to its left margin, with a page break on each side.
- The pagination resolves one context per page name, with the bands measured at that page's width, cuts each page at its own window height, and draws every page with its context's frame. Viewport units take the first page's area, so a document that opens with a cover sizes `100vh` to it.

This approximates css-page-3 §8.1: the group's ancestors keep the unnamed page's width, a box inside a group cannot open another, and the page pseudo-classes are parsed but never matched. The docs say so. Goldens `named-page-cover` and `named-page-inside` pin a cover that opens the document and one in the middle of it.

Stacked on #1625.

## Summary by CodeRabbit

- **New Features**
  - Added named-page support for PDF output using `page: <name>` and matching `@page <name>` rules.
  - Named pages support distinct sizes, margins, viewport-unit behavior, and header/footer bands.
  - Named-page sections begin on a separate page and return to default page styling afterward.

- **Bug Fixes**
  - Improved pagination, repeated content, and link placement across pages with different sizes or margins.

- **Documentation**
  - Added guidance on named-page properties, behavior, supported styling, and limitations.


kane50613:
Closing in favour of the option-level page bands in #1627. The `@page` parser and named-page contexts stay in this branch's history if a CSS-driven page geometry comes back.

PR #1627, opened 2026-09-14
PDF: override the header and footer on some pages

### LINKED ISSUE #1599: Support a footer/section that renders only on the last page, aligned to the bottom margin
### Pre-submission Checklist

- [x] I have searched [existing issues](https://github.com/kane50613/takumi/issues) and this feature has not been requested

### Problem Statement

render()'s footer option repeats on every page, which is great for page numbers or persistent notices, but there's no way to render content that appears only once, on the last page, pinned to the bottom margin (e.g., a closing "Thank you for your business" note, terms, or a signature block on an invoice).

Right now that kind of content has to be placed as regular body content, which flows in-line after whatever comes before it. If the preceding content doesn't happen to fill the page exactly, the note ends up wherever the flow lands — often near the top of the final page — instead of sitting flush against the bottom. Since Takumi paginates with Taffy rather than a browser, there's no flex/sticky/margin-top: auto trick that reaches across a page break to fix this from userland CSS.

### Proposed Solution

Add a lastPageFooter (or similar) render option, analogous to footer, that:

Renders once, only on the final page.
Is bottom-aligned within that page's margin area, the same way footer is bottom-aligned on every page.
Doesn't require the content author to know in advance how many pages the document will have.

### Proposed API / Usage

```javascript
const pdf = await render(doc, {
  size: "a4",
  footer: <PageNumberFooter />,       // every page
  lastPageFooter: <ClosingNote /> or <InvoiceTotal />,    // only the final page, bottom-aligned
});
```

### Alternatives Considered

- Using footer for the closing note anyway. Works, but forces the note onto every page, which isn't appropriate for one-off content like a final thank-you message that shouldn't repeat on page 1 of a 2-page invoice.
- Manually padding body content with a spacer <div> sized to fill remaining page height. Technically works today, but requires the caller to precompute how much vertical space is left on the page before the note, which means re-implementing part of Takumi's own layout/pagination logic in userland, and it breaks any time font size, margins, or content length change.
- A CSS @page:last selector or position: running() (Paged Media style). Would solve this more generally (and could handle first-page/last-page headers too), but is a much bigger CSS-engine feature than a single new render option, so it's listed here as a longer-term alternative rather than the immediate ask.

### Related Package

@takumi-rs/core

### Additional Context

_No response_

kane50613:
Fixes #1599

## Why
A footer that only the last page shows, or a header the cover leaves out, had no home: `header` and `footer` took one tree for every page.

## What
`pages` takes `{ first, last, odd, even }`, each overriding `header` or `footer` for the pages it covers. A page reads each field from `first` or `last`, then its parity, then the top-level option, and `false` stops the fall. Every tree a band can draw is measured and the automatic margin fits the tallest, so a taller last-page footer does not move the other pages' breaks.

```ts
type Band = Node | false;
type PageOverride = { header?: Band; footer?: Band };
type PageRules = { first?: PageOverride; last?: PageOverride; odd?: PageOverride; even?: PageOverride };
```

A band field takes three values and each means something: left out inherits, `false` draws nothing, a node draws itself. `header` and `footer` take a band at the top level too, so the same value works in both places.

`PageRules` rather than `PageVariants`, because nothing varies: the four keys select the pages a rule covers, which is what CSS calls a page rule. `PageOverride` keeps its name, since `@page` takes `size`, `margin` and `page-orientation` as well as its bands and the type has room to grow into them.

The `page-rules` golden pins a header that stays off page 1 and a last page with a two-line closing footer, and the test checks the page count matches a render with that footer on every page.

## Summary by CodeRabbit

- **New Features**
  - Added per-page header and footer rules for first, last, odd, and even pages.
  - Header and footer content can be disabled for selected pages using `false`.
  - Added precedence handling for overlapping rules, including single-page documents.
  - Page layout now accommodates varying header and footer sizes across pages.
  - Renamed page variants to page rules for clearer configuration.
  - Added configurable handling for missing glyphs: error, placeholder, or blank output.

- **Documentation**
  - Added guidance and examples covering page-specific headers, footers, precedence, page numbering, ranges, and automatic margin behavior.


kane50613:
Both findings were against the earlier shape, where `header` and `footer` each took either a node or a variants object. That branch is gone: `header` and `footer` are plain node inputs again, and the variants moved to `pages`, so there is no `isBandVariants` discriminator left to misclassify `{}`.

Unknown keys are rejected as the second finding asked, at both levels: `rejectUnknownKeys` runs over `pages` itself and over each `pages.<variant>` before any node resolution, so `pages.frist` and `pages.first.headr` throw instead of being dropped. The wasm side keeps `deny_unknown_fields` on the same two structs.

Issue #1541, opened 2026-09-03
Expose computed styles in measure()


benhid:
### Pre-submission Checklist

- [x] I have searched [existing issues](https://github.com/kane50613/takumi/issues) and this feature has not been requested

### Problem Statement

I'm using takumi to turn a slide written in HTML into an editable PowerPoint file (native text boxes, shapes and tables via PptxGenJS). Geometry comes from `measure()`; for the paint I currently re-implement the CSS cascade in JavaScript against the same stylesheet: selector matching, specificity, inheritance, `var()`, shorthands. That is ~250 lines that duplicate work the engine has already done, and it silently drifts whenever takumi's CSS support changes.

### Proposed Solution

Add an opt-in `includeStyles` option to `measure()`. When set, each `MeasuredNode` and each `MeasuredTextRun` carries a `style` object with the resolved values the renderer used for that box or run. Everything listed below is already computed inside the render pipeline, so this should be a serialisation change rather than new layout work.

Keeping it opt-in leaves the default payload unchanged:

```ts
interface MeasuredStyle {
  color?: string;            // "#14110F" or "rgba(...)"
  fontFamily?: string;       // the family actually used, e.g. "Georgia"
  fontSize?: number;         // px
  fontWeight?: number;       // 400, 700
  fontStyle?: "normal" | "italic" | "oblique";
  lineHeight?: number;       // px
  letterSpacing?: number;    // px
  textAlign?: string;
  textTransform?: string;
  backgroundColor?: string;
  backgroundImage?: string;  // raw value, e.g. "linear-gradient(...)"
  borderRadius?: number;     // px
  borderWidths?: [number, number, number, number];
  borderColors?: [string, string, string, string];
  boxShadow?: string;        // raw value
  opacity?: number;
  zIndex?: number;
}

interface MeasuredNode { /* existing fields */ style?: MeasuredStyle }
interface MeasuredTextRun {
  /* existing fields */
  style?: Pick<MeasuredStyle, "color" | "fontFamily" | "fontSize" | "fontWeight" | "fontStyle" | "letterSpacing">
}
```

### Proposed API / Usage

```js
import { Renderer } from "@takumi-rs/core";
import { fromHtml } from "@takumi-rs/helpers/html";

const { node, css } = fromHtml(`
  <style>.big { font-family: Georgia; font-size: 120px; color: #B3261E; font-weight: 700 }</style>
  <div style="width: 1280px; height: 720px; padding: 88px; background: #F7F3EC">
[...]

PR #1632, opened 2026-09-14
Core: report each text run's ascent from measure()

benhid:
A run's `height` is `ascent + descent` with no way to recover the split, so a consumer handing the text to another layout engine could not find the baseline. Runs now report `ascent`, and `y + ascent` is the baseline: the same value for every run on a line, whatever their font sizes.

It is geometry rather than paint, so it is always present rather than behind an option.

`MeasuredTextRun` gains a field, which breaks a downstream struct literal in Rust. It is not `#[non_exhaustive]` because the repo's own tests construct it.

## Tests

`test_measure_run_ascent_locates_the_shared_baseline` puts a 48px run and a 16px run on one line and asserts their ascents differ while `y + ascent` agrees. The measure fixture comparison asserts `ascent` now too.

## Verification

```
cargo test -p takumi-core                    1039 pass
cargo test -p takumi                         goldens unchanged
cargo check -p takumi-napi
cargo check -p takumi-wasm --target wasm32-unknown-unknown
cargo clippy --locked --all-targets --all-features -- -D warnings
cargo fmt --all · bun run lint
```

## Summary by CodeRabbit

* **New Features**
  * Measured text runs now expose an `ascent` value indicating the distance from the run’s top edge to its baseline.
  * Consumers can determine baseline placement using `y + ascent`, including for runs with different font sizes.

* **Documentation**
  * Updated measurement API documentation with the new field and baseline guidance.

* **Tests**
  * Added coverage verifying ascent values and shared baseline alignment.


kane50613:
I'm curious why need to hand text to another layout engine, whats the use case?

benhid:
> I'm curious why need to hand text to another layout engine, whats the use case?

The use case is #1541: an HTML deck turned into an editable `.pptx` (takumi's `measure()` does the layout, PptxGenJS writes the deck), where each paragraph becomes a real PowerPoint text box.

PowerPoint has no `line-height`. It builds each line from the font's metrics, so to place a text box I need to know where the baseline sits inside the box. `height` is `ascent + descent`, so the baseline is not recoverable from it. I was working around this with a hand-measured table, one number per font family, which drifted for any family not in it.

benhid:
Closing in favour of #1642

PR #1652, opened 2026-09-16
PDF: scale gradient transforms to the page unit

benhid:
Every gradient in a PDF draws 4/3 too large. A linear ramp runs off its box, so its final colour
never appears, and radial and conic ones also land off-centre. The same markup rasterizes
correctly, so only PDF output is affected.

```html
<div style="width: 300px; height: 120px; background-image: linear-gradient(to right, #ff5f6d, #3a1c71);"></div>
```

Top row is the raster backend, middle is the PDF today, bottom is this PR:

comparison-gradients

## Cause

The emitter works in css pixels and converts to points by pushing a scale onto the surface, so
every path it draws inherits the conversion. A shading pattern does not: PDF 32000-1 8.7.3.1 maps
a pattern matrix to the default space of the page, and a tiled background opens a stream of its
own whose transform starts at identity. The gradient is built in pixels and read as points, and
what is left over is `1 / PT_PER_PX`, or 4/3.

The geometry itself is correct. `resolve_geometry` produces the same values the raster backend
paints correctly, so this is a unit error, not a maths one.

## Why the conversion is conditional

A gradient painted straight onto the surface, or into the stream of a mask, already inherits the
page scale, and converting it again would leave it at 0.5625. So `in_page_pattern` travels down
from the call site that opens the pattern stream, and only a gradient whose pattern resolves
against the page carries the conversion.

Mean per-channel error against the raster backend on the same markup:

| case | before | after |
| ---- | ------ | ----- |
| `gradients` fixture | 12.24 | **0.16** |
| `mask-image` fixture | 1.71 | **0.36** |
| `background-clip: text` | 0.58 | **0.58** |

Conic gradients were affected too, for the same reason: their centre is in pixels like the
radial's.

## An alternative

The other way to fix this is to make a pattern's nested stream inherit the page transform inside
krilla, which removes the flag and corrects all three routes at once. It is the cleaner statement
of the same fact, and it reaches anything else that opens a nested stream. I kept the change in
the emitter to hold the blast radius to gradients, but I am happy to redo it that way if you
prefer.

## Tests

`a_gradient_axis_is_measured_in_page_units` renders a 300x120 box, reads the shading's `/Coords`
and its pattern's `/Matrix` out of the PDF, composes them, and asserts the axis measures 225 pt.
Reverting the fix fails it with `gradient axis is 300 pt, expected 225`.

The existing `gradients` fixture could not catch this. `run_pdf_fixture` checks determinism and
the `%PDF-` header and writes the golden, but compares against no expectation, and wrong gradient
geometry does not announce itself in a binary diff.

## Goldens

Six PDF goldens move, and the move is the fix rather than a side effect: `gradients`,
`repeating-gradients`, `mask-image`, `clip-path-shapes`, `background-placement` and
`overflow-clip`. No raster golden moves.

## Verification

```
cargo test -p takumi-pdf                     113 pass
cargo clippy --locked --all-targets --all-features -- -D warnings
cargo fmt --all · bun run lint
```

Correctness was checked against takumi's own raster backend on identical markup, with poppler used
only to rasterize for the measurements.

`background_layer` reaches eight arguments and takes `#[allow(clippy::too_many_arguments)]`,
matching `tiled_layer` and three other functions in the same file. Happy to group `x` and `y` into
a tuple instead, the way `rect_at` and `anchor` already are.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected gradient scaling in PDF output so gradients align with page units.
  * Fixed oversized or off-center radial and conic gradients.
  * Improved gradient rendering across backgrounds, masks, patterns, and clipped fills.

* **Tests**
  * Added coverage verifying that gradient dimensions are accurately converted from CSS pixels to PDF points.

* **Documentation**
  * Added a patch note describing the gradient rendering corrections.


kane50613:
I measured the rendered output rather than trusting the goldens, and pushed the assertions that came out of it (33d39e7).

The fix is correct. Three independent checks:

**The pattern matrix itself, which no renderer gets a say in.** A 110 css px box is 82.5 pt, so its centre is 41.25 pt in. After the fix the radial shading's matrix is `scale 0.75, translate (41.25, 41.25)` and the circle centre maps to exactly that; the radius maps to 58.34 pt, which is the farthest corner (55·√2 px). Before, the matrix was `scale 1.0, translate (55, 55)`, putting the centre 13.75 pt down and right and the radius 4/3 too long.

**Poppler at 96 dpi.** Sampling the radial box at equal distances left, right, up and down from its centre returns byte-identical colours at every radius, and the pixel closest to `#fddb92` sits exactly on the centre. The conic's hue ring matches the stop list at every quarter turn.

**Against takumi's own raster backend**, same markup, per-channel mean error:

| | before | after |
| --- | --- | --- |
| linear | 11.06 | 0.15 |
| radial | 18.94 | 0.38 |
| conic | 41.80 | 0.41 |

The measured radial offset before the fix was +18 px on both axes, which is the 13.75 pt the matrix predicted.

Worth saying why the new tests exist: the byte goldens are checked, CI regenerates them and fails on any drift, so I was wrong to call them unverified earlier. What they cannot do is tell a wrong starting point from a right one, and the golden that drew every gradient 4/3 too large passed that check for as long as it stayed wrong. `a_gradient_axis_is_measured_in_page_units` covered the linear case; the two new tests do the same for the radial centre and reach and for the conic scale. Reverting the fix now fails with `gradient centre is (60, 60) pt, expected (45, 45)` instead of a binary diff.

PR #1664, opened 2026-09-19
Raster: scan mask alpha rows without per-pixel branches

kane50613:
## Why

Box shadows and backdrop filters attenuate a full alpha plane by a mask and scan the canvas for the shadow's bounds. Both loops branched on every pixel, which kept LLVM from vectorizing them.

## What

The attenuation multiplies every byte, since `fast_div_255(d * 255)` is `d`. The bounds scan finds each row's first and last visible pixel. Output is byte-identical. On the example 4K shadow scene this saves about 9%, and on the backdrop-blur scene about 11%.

## Summary by CodeRabbit

- **Performance Improvements**
  - Box shadows and backdrop filters now render faster.
  - Alpha masking and visible-area detection are processed more efficiently, improving rendering performance without changing visual results.
- **Bug Fixes**
  - Preserved existing masking behavior, including correct handling of transparent regions and overlapping areas.
  - Maintained accurate shadow and filter bounds across edge cases, including single-pixel, corner, and fully transparent masks.


kane50613:
@coderabbitai review

PR #1665, opened 2026-09-19
Raster: fill oblique linear gradients four rows at a time

kane50613:
## Why

A linear gradient at an angle projected, rounded, and looked up every pixel through a serial per-row recurrence. On a 1200x630 card it was the largest single cost after the demultiply pass.

## What

Opaque, non-repeating, undithered gradients fill four rows per pass. Each row keeps its own `projection += dir_x` recurrence, so the result stays byte-identical, and `simd::linear_lut_indices` turns the four projections into LUT indices with NEON, SSE2, or wasm simd128. The SSE2 and simd128 backends emulate round-half-away-from-zero with `trunc + (frac >= 0.5)`, checked against the scalar version over 200k inputs. The example card scene is about 20% faster at both 1200x630 and 4K.

## Summary by CodeRabbit

* **Performance**
  * Improved rendering performance for opaque, non-repeating angled linear gradients by processing multiple rows at once.
  * Added optimized gradient calculations across supported hardware acceleration paths, with a compatible fallback when unavailable.

* **Rendering**
  * Improved consistency of angled gradient output across different tile sizes and orientations.
  * Preserved correct handling for repeating, translucent, and dithered gradients.


kane50613:
@coderabbitai review

PR #1667, opened 2026-09-19
Raster: run the RGBA vertical blur pass over flat bytes

kane50613:
## Why

The RGBA vertical box blur kept `[u32; 4]` sums per pixel, and LLVM vectorized that layout with a transpose on every step.

## What

The vertical pass treats each row as `width * 4` independent bytes and reuses the alpha kernel. That is byte-identical, because every channel slides the same window down its own column. `effects/backdrop_blur_cards` is about 12% faster.

## Summary by CodeRabbit

  * Improved the speed of vertical RGBA backdrop blur processing.

* **Bug Fixes**
  * Improved consistency of vertical RGBA blur results across different blur radii and image dimensions.


kane50613:
@coderabbitai review

